PR #12, opened 2020-11-03
Apply Triton code format

deadeyegoodwin:


deadeyegoodwin:
@szalpal  Triton uses clang-format to format source. You should apply this format as part of every PR. The format script itself is in the common repo so you will need to check that out.

$ python3 ../common/tools/format.py src

szalpal (approved):

PR #13, opened 2020-11-04
Capture time statistics & docs update

szalpal:
TODO:
- [x] Test if solution works
- [x] Wait for https://github.com/triton-inference-server/common/tree/davidg-format to be merged
- [x] Reformat code with `clang-format`

A few things happen in this PR:
- DALI Backend from now on will gather time statistics and report them via clients
- Small description of the docker image, handy for development
- Code base has been formatted according to triton-inference-server practise
- Updating supplied client to `tritonserver:20.10-client-py3`
- Small fixes due to gcc upgrade (proper to ubuntu20)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:20.09-py3
+ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:20.10-py3
 FROM ${BASE_IMAGE} as builder
 
 RUN apt-get update && \
@@ -37,10 +37,13 @@ RUN add-apt-repository ppa:deadsnakes/ppa && \
               python3.8        \
               python3-pip      \
               libboost-all-dev \
-              rapidjson-dev
+              rapidjson-dev    \
+              gdb
 
 # pip version in apt packages is ancient - we need to update it
 RUN pip3 install -U pip
+
+WORKDIR /opt
 
 # CMake
 RUN CMAKE_VERSION=3.17 && \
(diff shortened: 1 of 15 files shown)


Dockerfile
```diff
@@ -37,11 +37,14 @@ RUN add-apt-repository ppa:deadsnakes/ppa && \
               python3.8        \
               python3-pip      \
               libboost-all-dev \
-              rapidjson-dev
+              rapidjson-dev    \
+              gdb
```
JanuszL, line 41: ?

src/dali_backend.cc
```diff
@@ -426,23 +429,29 @@ TRITONBACKEND_ModelInstanceExecute(
...
       error = TRITONSERVER_ErrorNew(
-          TRITONSERVER_Error_Code::TRITONSERVER_ERROR_UNKNOWN,
-          "Unknown DALI Backend error");
+              TRITONSERVER_Error_Code::TRITONSERVER_ERROR_UNKNOWN,
+              "Unknown DALI Backend error");
     }
 
+    exec_end_ns = detail::capture_time();
+
+    RETURN_IF_ERROR(TRITONBACKEND_ModelInstanceReportStatistics(
```
deadeyegoodwin, line 446: You should also report the overall batch statistics, like here: https://github.com/triton-inference-server/onnxruntime_backend/blob/main/src/onnxruntime.cc#L1135

deadeyegoodwin (changes requested):

src/dali_backend.cc
szalpal: Done

src/dali_backend.cc
```diff
@@ -396,19 +398,25 @@ TRITONBACKEND_ModelInstanceExecute(
...
+      batch_compute_start_ns = -1, batch_compute_end_ns = -1;
+  int total_batch_size = 0;
   DaliModelInstance* instance_state;
   RETURN_IF_ERROR(TRITONBACKEND_ModelInstanceState(
       instance, reinterpret_cast<void**>(&instance_state)));
   std::vector<TRITONBACKEND_Request*> requests(reqs, reqs + request_count);
   std::vector<TRITONBACKEND_Response*> responses(request_count);
 
   for (size_t i = 0; i < responses.size(); i++) {
+    exec_start_ns = detail::capture_time();
```
deadeyegoodwin, line 412: You should move the decl of error = nullptr to here. You need to set 'error' = nullptr for each request. In general need to think about what happens if one or more requests have error.

src/dali_backend.cc
```diff
@@ -436,14 +444,30 @@ TRITONBACKEND_ModelInstanceExecute(
           "Unknown DALI Backend error");
     }
 
+    exec_end_ns = detail::capture_time();
+
+    RETURN_IF_ERROR(TRITONBACKEND_ModelInstanceReportStatistics(
```
deadeyegoodwin, line 449: Can't return from this function until all responses have been sent. See identity example. This function can return error only if no responses are sent. But as soon as any response is sent (eigher success or error) then all responses must be sent and the function must return success.

So, the one RETURN_IF_ERROR in the loop above and the three RETURN_IF_ERROR here must be changed to LOG_IF_ERROR.

src/dali_backend.h
```diff
@@ -164,20 +173,34 @@ AllocateOutputs(
...
-void
+
+struct RequestMeta {
+  uint64_t compute_start_ns, compute_end_ns;
+  int batch_size;
+};
+
+
+RequestMeta
 ProcessRequest(
```
deadeyegoodwin, line 184: Why is this function in the header?

src/dali_backend.cc
```diff
@@ -21,7 +21,9 @@
 // SOFTWARE.
 
 #include "src/dali_backend.h"
+
```
deadeyegoodwin, line 24: In triton coding convention we put a blank line here after the corresponding header file (if any)....

src/dali_backend.cc
```diff
@@ -21,7 +21,9 @@
...
 #include <memory>
+
```
deadeyegoodwin, line 26: But we don't put this blank line. We just let the formatter order all the includes together.

deadeyegoodwin (changes requested):

src/dali_backend.cc
szalpal: This (and the one below) is actually what formatter gave me. I tried to rerun it multiple times, but it still insists on having those 2 blank lines...

src/dali_backend.h
szalpal: I've put the dali_backend-specific functions inside a header to have clearer separation of the custom routines and backend's boiler-plate.

src/dali_backend.cc
szalpal: Done. I've left `RETURN_IF_ERROR` only for `ModelInstanceState` call.

src/dali_backend.cc
szalpal: Done

src/dali_backend.cc
```diff
@@ -436,14 +444,30 @@ TRITONBACKEND_ModelInstanceExecute(
...
         requests[i],
         TRITONSERVER_RequestReleaseFlag::TRITONSERVER_REQUEST_RELEASE_ALL));
+
+    total_batch_size += request_meta.batch_size;
+    batch_exec_end_ns = exec_end_ns;
+    batch_compute_end_ns = request_meta.compute_end_ns;
   }
+
+  RETURN_IF_ERROR(TRITONBACKEND_ModelInstanceReportBatchStatistics(
+      instance, total_batch_size, batch_exec_start_ns, batch_compute_start_ns,
```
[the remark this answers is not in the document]
szalpal: Yeah, I've missed it.
Fixed setting the `batch_...`. At one point I needed to use the ternary instead of a simple if, cause the compiler was complaining about `batch_compute_start_ns may be used uninitialized in this function`.

src/dali_backend.cc
deadeyegoodwin: Did you run format.py with python or python3?

PR #15, opened 2021-01-08
Init TRITONBACKEND_OutputBuffer arguments that also serve as input

GuanLuo:
mem_type and mem_type_id specifies the preferred value when TRITONBACKEND_OutputBuffer reads them as input

### DIFF
--- a/src/dali_backend.h
+++ b/src/dali_backend.h
@@ -158,8 +158,8 @@ AllocateOutputs(
         response, &triton_output, name, to_triton(snt.type),
         output_shape.data(), output_shape.size());
     void* buffer;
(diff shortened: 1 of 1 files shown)


szalpal (approved):
I'm not sure, what does "memory type id" mean. If it's index in the `TRITONSERVER_MemoryType` enum, it should rather be `memid = 1` for `TRITONSERVER_MEMORY_CPU_PINNED`, right? @deadeyegoodwin , could you help here?

Otherwise LGTM

deadeyegoodwin:
@szalpal The memory type id is a memory-type specific identifier. For GPU memory type it indicates the GPU device. For CPU memory types it should always be 0.

PR #23, opened 2021-02-09
Readme update: info about latest dali_backend version

szalpal:


README.md
```diff
@@ -91,12 +99,14 @@ group for DALI model.
...
 ## How to build?
 
-### Docker build
+### Docker build <a name="docker_build"></a>
 Building DALI Backend with docker is as simple as:
 
     git clone --recursive https://github.com/triton-inference-server/dali_backend.git
     cd dali_backend 
-    docker build .
+    docker build -t tritonserver:dali-latest .
```
deadeyegoodwin, line 107: I notice in the dockerfile you build cmake. You could just install it using apt like we do here: https://github.com/triton-inference-server/server/blob/master/build.py#L467

deadeyegoodwin (approved):

README.md
JanuszL: The default one in Ubuntu 18.04 was too old for DALI (3.10 vs 3.13). Maybe one from Ubuntu 20.04 will do. Is Triton based already on Ubuntu 20.04?

PR #24, opened 2021-02-10
Utilize variable batch size in DALI pipelines.

banasraf:
This PR removes the `PipelineGroup` abstraction and makes use of variable batch size in DALI external source. 


szalpal:
Waiting for `DALI v0.31`

src/dali_executor/dali_pipeline.cc
```diff
@@ -26,14 +26,27 @@
...
+{
+  auto dim = shape.ndim;
+  auto batch_size = shape.num_samples();
+  std::vector<int64_t> result(dim * batch_size);
+  for (int64_t s = 0; s < batch_size; ++s) {
+    for (int64_t d = 0; d < dim; ++d) {
+      result[s * dim + d] = shape[s][d];
+    }
+  }
+  return result;
```
mzient, line 40:
```suggestion
  return shape.shapes;
```
Note that this doesn't work for 0D tensor lists.
Also - do you really need to make this copy?

src/dali_executor/dali_pipeline.cc
```diff
@@ -67,12 +80,23 @@ DaliPipeline::SetInput(
...
 }
 
 
+void
+DaliPipeline::SetInput(
+    const void* ptr, const char* name, device_type_t source_device,
+    dali_data_type_t data_type, TensorListShape<> input_shape) {
+  auto shape_list = dali_to_list_shape(input_shape);
+  const int n_dim = input_shape.ndim;
+  SetInput(ptr, name, source_device, data_type, make_span(shape_list), n_dim);
```
mzient, line 97:
```suggestion
    dali_data_type_t data_type, const TensorListShape<> &input_shape) {
  SetInput(ptr, name, source_device, data_type, make_span(input_shape.shapes), input_shape.sample_dim());
``````

File: src/dali_executor/dali_executor.cc [hunk 3]
```diff
@@ -26,35 +26,45 @@
...
+  int batch_size = inputs[0].shape.num_samples();
+  for (size_t i = 1; i < inputs.size(); ++i) {
+    assert(
+        inputs[i].shape.num_samples() == batch_size &&
+        "All inputs should have equal batch size.");
+  }
+  for (auto& inp : inputs) {
+    assert(
+        inp.shape.num_elements() * dali_type_size(inp.type) <=
+        inp.buffer.size());
```
mzient, line 43: Is this line break really required? This looks weird.

src/dali_executor/test/test_utils.h
```diff
@@ -75,6 +75,7 @@ RandomInput(
   dscr.name = name;
   dscr.shape = shape;
   dscr.type = dali_data_type<T>();
+  dscr.device = device_type_t::CPU;
```
mzient, line 78: Remove? It's set again two lines below.

banasraf: done

src/dali_executor/dali_executor.cc
banasraf: We use clang-format used across all of the triton-server code.

src/dali_executor/dali_pipeline.cc
banasraf: done

src/dali_executor/dali_pipeline.cc
banasraf: Nope the copy was not needed. Removed this function

src/dali_executor/pipeline_pool.h
```diff
@@ -67,7 +67,7 @@ struct hash<triton::backend::dali::PipelineDescr> {
   size_t operator()(
       triton::backend::dali::PipelineDescr const& pk) const noexcept
   {
-    return pk.serialized_pipeline_hash ^ static_cast<size_t>(pk.batch_size);
+    return pk.serialized_pipeline_hash ^ static_cast<size_t>(pk.max_batch_size);
```
mzient, line 70:
```suggestion
    return pk.serialized_pipeline_hash + static_cast<size_t>(pk.max_batch_size);
```
+ is better than XOR, because XORing with the same number multiple times just flips between two values.

src/dali_executor/dali_pipeline.h
```diff
@@ -110,10 +110,14 @@ class DaliPipeline {
...
 
+  void SetInput(
+      const void* ptr, const char* name, device_type_t source_device,
+      dali_data_type_t data_type, TensorListShape<> input_shape);
+
   void PutOutput(
       void* destination, int output_idx, device_type_t destination_device);
 
-  const int batch_size;
+  const int max_batch_size;
```
szalpal, line 120: Since you're changing it already, could you change it to `const in kMaxBatchSize` or `int max_batch_size_`?

klecki, line 120: This is movable (but it doesn't have a move assignment), and probably one reason of this is the const member. Some people discourage using them.

Also, the default constructor (or rather one arg constructor with default argument max_batch_size=0 creates a rather useless pipeline).

qa/multi_input/multi_input_client.py
```diff
@@ -55,18 +57,19 @@ def array_from_list(arrays):
     return np.stack(arrays)
 
 
-def batcher(dataset, batch_size, n_iterations=-1):
+def batcher(dataset, max_batch_size, n_iterations=-1):
```
klecki, line 60: I would suggest creating test_utils ad some point to have the batcher() and randomly_shaped_samples() there.

klecki (commented):
Some testing interface nitpick, and suggestion about the DaliPipeline wrapper, otherwise ok.

mzient (dismissed):

src/dali_executor/pipeline_pool.h
banasraf: done

src/dali_executor/dali_pipeline.h
banasraf: renamed to max_batch_size_

banasraf: I changed the default constructor to create empty handle that does not create stream, nor initialize DALI

banasraf: Also added move assign

szalpal (dismissed):

klecki (approved):

szalpal (approved):

PR #27, opened 2021-02-17
Add test for pipeline restarting

banasraf:
Add a test case for recovering from pipeline failure. 


src/dali_executor/executor.test.cc
```diff
@@ -86,4 +86,56 @@ TEST_CASE("Scaling Pipeline")
...
+    auto output = executor.Run(std::vector<IODescr<false>>({input}));
+    std::vector<float> output_buffer(output[0].shape.num_elements());
+    std::vector<IODescr<false>> output_vec(1);
+    auto& outdesc = output_vec[0];
+    outdesc.device = device_type_t::GPU;
+    outdesc.device_id = 0;
+    outdesc.buffer = make_span(
+        (char*)output_buffer.data(),
+        output_buffer.size() * sizeof(decltype(output_buffer)::size_type));
+    executor.PutOutputs(output_vec);
```
JanuszL, line 114: Can we validate the output anyhow?

banasraf: Hard to say. I could validate the shape but I'm not sure about contents as it's Resnet pipeline run on some random small image. The purpose of this test was rather to check for exceptions

JanuszL: If the image is the same, I guess we can expect rather quite accurate Top1 or at least Top5.

banasraf: The image was single (0, 0, 0) pixel, so now I verify the output values.

JanuszL (approved):

PR #47, opened 2021-04-16
Add num_threads option to config parameters.

banasraf:
Refactoring: 
I removed the PipelinePool to simplify the executor.

New feature:
I added a  ModelParameters class do dali_backend.cc that parses "parameters" map to extract num_threads.


src/dali_executor/utils/utils.h
```diff
@@ -60,6 +60,13 @@ array_shape(TensorListShape<ndims> tls)
   return ret;
 }
 
+template <typename T>
+T string_as(const std::string &str);
```
JanuszL, line 64: How about `from_string`?

src/dali_executor/dali_executor.h
```diff
@@ -42,9 +42,12 @@ struct shape_and_type_t {
...
  public:
   DaliExecutor(
-      std::string serialized_pipeline, int max_batch_size, int device_id)
+      std::string serialized_pipeline, int max_batch_size, int num_threads, int device_id)
       : serialized_pipeline_(std::move(serialized_pipeline)),
-        device_id_(device_id), max_batch_size_(max_batch_size)
+        max_batch_size_(max_batch_size),
+        num_threads_(num_threads),
+        device_id_(device_id),
+        pipeline_(serialized_pipeline_, max_batch_size, num_threads, device_id)
```
JanuszL, line 50: What if `serialized_pipeline_` if invalid and will throw?

src/dali_executor/dali_pipeline.h
```diff
@@ -71,14 +71,13 @@ class DaliPipeline {
...
-      int device_id = -1, int bytes_per_sample_hint = 0, int num_threads = -1,
-      int seed = -1)
+      int num_threads = -1, int device_id = -1)
   {
     InitDali();
     InitStream();
     daliCreatePipeline(
         &handle_, serialized_pipeline.c_str(), serialized_pipeline.length(),
-        max_batch_size, num_threads, device_id, 0, 1, 666, 666, 0);
+        max_batch_size, num_threads, device_id, 0, 1, 0, 0, 0);
```
JanuszL, line 80: What this 666->0 changes?

src/dali_backend.cc
```diff
@@ -26,9 +26,43 @@
...
+  template <typename T>
+  void GetMember(
+      common::TritonJson::Value &params,
+      const char* key,
+      T *value)
+  {
+    if (params.Find(key)) {
+      common::TritonJson::Value param;
+      TRITON_CALL_GUARD(
+        params.MemberAsObject(key, &param));
```
JanuszL, line 51: What is that for?

src/dali_backend.cc
```diff
@@ -26,9 +26,43 @@
...
+      std::string string_value;
+      TRITON_CALL_GUARD(
+        param.MemberAsString("string_value", &string_value));
```
JanuszL, line 54: Maybe `"string_value"` should be an argument of this `GetMember` function? Or do you plan to name it like that no matter what?

src/dali_backend.cc
```diff
@@ -26,9 +26,43 @@
...
+      num_threads = string_as<int>(string_value);
```
JanuszL, line 55:
```suggestion
      *v = string_as<T>(string_value);
```

src/dali_backend.cc
```diff
@@ -26,9 +26,43 @@
...
+    model_config.MemberAsObject("parameters", &params);
+    GetMember(params, num_threads_key, &num_threads);
+  }
+
+ private:
+  template <typename T>
+  void GetMember(
+      common::TritonJson::Value &params,
+      const char* key,
+      T *value)
```
JanuszL, line 46:
```suggestion
      T &value)
```
?

src/dali_backend.cc
```diff
@@ -26,9 +26,43 @@
...
+    common::TritonJson::Value params;
+    model_config.MemberAsObject("parameters", &params);
+    GetMember(params, num_threads_key, &num_threads);
+  }
+
+ private:
+  template <typename T>
+  void GetMember(
+      common::TritonJson::Value &params,
+      const char* key,
```
JanuszL, line 45:
```suggestion
      const char &key,
```

src/dali_executor/dali_pipeline.h
banasraf: Those are the separate prefetch queue depths, so they don't matter here. I changed it because because I don't see a reason for it to be 666

src/dali_backend.cc
banasraf: The whole "parameters" section in this config is defined in Protobuff as `map<string, ModelParameter>` and ModelParameters is: 
```
message ModelParameter
{
  string string_value = 1;
}
```
So no matter the specific parameter there will always be this `string_value` field.

src/dali_backend.cc
banasraf: it's a c style string, so it has to be pointer, right?

src/dali_executor/dali_pipeline.h
JanuszL: ok

src/dali_backend.cc
JanuszL: Sure

src/dali_backend.cc
banasraf: fixed

src/dali_backend.cc
banasraf: done.
I wanted to make the signature similar to the GetMemberAs... from triton API, but there is no need for that actually.

src/dali_executor/utils/utils.h
banasraf: done

src/dali_backend.cc
JanuszL: The question is if it has to be a c style string? Just asking.

banasraf: A static const std::string has to be initialized outside of the class, and I thought it's cleaner when all the (future) key strings are listed inside this class.

JanuszL: OK

src/dali_executor/dali_executor.h
banasraf: It's caught and handled when instantiating the model
https://github.com/triton-inference-server/dali_backend/blob/2b0f5139d72e093708eb42dffb0b7f940c7546ab/src/dali_backend.cc#L153

src/dali_backend.cc
```diff
@@ -26,9 +26,43 @@
...
+      T &value)
+  {
+    if (params.Find(key)) {
+      common::TritonJson::Value param;
+      TRITON_CALL_GUARD(
+        params.MemberAsObject(key, &param));
+      std::string string_value;
+      TRITON_CALL_GUARD(
+        param.MemberAsString("string_value", &string_value));
+      value = string_as<T>(string_value);
```
JanuszL, line 55:
```suggestion
      value = from_string<T>(string_value);
```

src/dali_executor/dali_executor.h
JanuszL: OK

src/dali_backend.cc
banasraf: fixed.
and compiled this time...

JanuszL (dismissed):

src/dali_executor/dali_executor.h
```diff
@@ -42,9 +42,12 @@ struct shape_and_type_t {
 class DaliExecutor {
  public:
   DaliExecutor(
-      std::string serialized_pipeline, int max_batch_size, int device_id)
+      std::string serialized_pipeline, int max_batch_size, int num_threads, int device_id)
       : serialized_pipeline_(std::move(serialized_pipeline)),
-        device_id_(device_id), max_batch_size_(max_batch_size)
+        max_batch_size_(max_batch_size),
```
szalpal, line 47: If currently there's 1-1 relation `DaliExecutor` to `DaliPipeline`, do we need `serialized_pipeline_` here? I'd consider designing it so that `DaliExecutor` just gets any `DaliPipeline`. After all, the executor doesn't have to know at all what's inside `DaliPipeline`, right? Just needs to run something through it

src/dali_executor/dali_executor.cc
```diff
@@ -50,21 +49,20 @@ template <bool owns>
...
   try {
-    pipeline.Run();
-    pipeline.Output();
+    pipeline_.Run();
+    pipeline_.Output();
   }
   catch (std::runtime_error& e) {
-    pipeline_pool_.Remove(serialized_pipeline_, max_batch_size_, device_id_);
+    pipeline_ = DaliPipeline(serialized_pipeline_, max_batch_size_,
+                             num_threads_, device_id_);
```
szalpal, line 59: Maybe it would be worth to consider allowing `DaliPipeline` to reset itself? `serialized_pipeline_` shouldn't change after reset, so something like `pipeline_.Reset()` should suffice

src/dali_backend.cc
```diff
@@ -26,9 +26,43 @@
 
 #include "triton/backend/backend_model.h"
 #include "triton/backend/backend_model_instance.h"
+#include "src/dali_executor/utils/utils.h"
 
 namespace triton { namespace backend { namespace dali {
 
+struct ModelParameters {
+  explicit ModelParameters(common::TritonJson::Value &model_config)
+  {
```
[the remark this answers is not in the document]
banasraf: Added `GetParam` and `GetNumThreads` shorthand.

src/dali_executor/dali_executor.h
banasraf: I moved the creation of pipeline out of executor and moved Resetting logic to pipeline.

JanuszL (dismissed):

src/dali_executor/dali_executor.cc
banasraf: done

src/dali_backend.cc
```diff
@@ -26,9 +26,52 @@
...
+    model_config.MemberAsObject("parameters", &params_);
+  }
+
+  /**
+   * Rerurn a value of a parameter with a given `key`
+   * or `def` if the parameter is not present.
+   */
+  template <typename T>
+  T GetParam(const std::string &key, const T &def) {
+    T result = def;
```
szalpal, line 45: How about?
```suggestion
  template <typename T>
  T GetParam(const std::string &key, const T &def = T()) {
    T result = def;
```

src/dali_executor/dali_pipeline.h
```diff
@@ -54,6 +54,10 @@ class DaliPipeline {
     if (this != &dp) {
       ReleasePipeline();
       ReleaseStream();
+      serialized_pipeline_ = std::move(dp.serialized_pipeline_);
```
szalpal, line 57: Please double check, if moving like this requires assigning default value `dp.serialized_pipeline_={}`

src/dali_backend.cc
```diff
@@ -145,9 +192,12 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
   {
-    dali_executor_ = std::make_unique<DaliExecutor>(
-        dali_model_->GetModelProvider().GetModel(), dali_model_->MaxBatchSize(),
-        device_id_);
+    auto serialized_pipeline = dali_model_->GetModelProvider().GetModel();
+    auto max_batch_size = dali_model_->MaxBatchSize();
+    auto num_threads = dali_model_->GetModelParamters().GetNumThreads();
+    DaliPipeline pipeline(serialized_pipeline, max_batch_size,
+                          num_threads, device_id_);
+    dali_executor_ = std::make_unique<DaliExecutor>(std::move(pipeline));
```
szalpal, line 200: Is this possible?
```c++
dali_executor_ = std::make_unique<DaliExecutor>({serialized_pipeline, max_batch_size, num_threads, device_id_});
```
Always one less call to `std::move` ;)

szalpal (approved):

src/dali_executor/dali_pipeline.h
banasraf: It doesn't

src/dali_backend.cc
banasraf: I would have to write `DaliPipeline{.....}`. It seems too long for me

src/dali_backend.cc
banasraf: done

szalpal (approved):

PR #50, opened 2021-04-22
Auto-format code

banasraf:
Added formatting script and auto-formatted the codebase


klecki (dismissed):

szalpal (commented):
Generally everything is OK, however I'm not sure if adding the `format.py` script is a good idea. Since it's from `common` repo, it's maintained there and that would mean we don't need to do this ourselves :) I don't oppose having such script in `dali_backend`, however maybe it would be better to have it when we would enforce formatting and linting in the CI?

banasraf:
> Generally everything is OK, however I'm not sure if adding the `format.py` script is a good idea. Since it's from `common` repo, it's maintained there and that would mean we don't need to do this ourselves :) I don't oppose having such script in `dali_backend`, however maybe it would be better to have it when we would enforce formatting and linting in the CI?

I removed the script.

szalpal (approved):

PR #53, opened 2021-04-28
Support GPU inputs. 

banasraf:
The PR enables DALI backend to accept inputs on GPU. 

Most of the PR is a necessary refactoring.
* `utlis/triton.h` was created to simplify using Triton API.
* `IODescr` was simplified to only be a non-owning descriptor that supports chunked memory. Owning descriptor was replaced with `IOBuffer`.
* `DaliExecutor` now handles copying input buffers if it's necessary and uses `ThreadPool` to do it.

Testing:
* C++ tests for `IOBuffer`
* Extended C++ tests for `DaliExecutor`
* Integration tests in L0_DALI_GPU_ensemble
    client.py sends requests to a Triton server instance that runs models defined in model_repository and it validates the results.

JIRA task: DALI-1478


src/utils/triton.h
```diff
@@ -0,0 +1,170 @@
...
+inline TRITONSERVER_DataType to_triton(dali_data_type_t t) {
+  assert(-1 <= t && t <= 11);
+  // Use the trick, that TRITONSERVER_DataType and dali_data_type_t
+  // types have more-or-less the same order, with few exceptions
+  if (t == -1)
+    return static_cast<TRITONSERVER_DataType>(0);
+  if (t == 11)
+    return static_cast<TRITONSERVER_DataType>(1);
+  return static_cast<TRITONSERVER_DataType>(t + 2);
+}
```
banasraf, line 77: Moved from dali_backend.h

src/dali_backend.h
```diff
@@ -144,42 +48,42 @@ std::vector<IODescr<true>> GenerateInputs(TRITONBACKEND_Request* request) {
...
-    void* buffer;
-    TRITONSERVER_MemoryType memtype = TRITONSERVER_MEMORY_GPU;
+    TRITONSERVER_MemoryType memtype = to_triton(out_info.device);
     int64_t memid = 0;
-    auto buffer_byte_size =
-        std::accumulate(output_shape.begin(), output_shape.end(), 1, std::multiplies<int>()) *
-        TRITONSERVER_DataTypeByteSize(to_triton(snt.type));
+    auto t_size = TRITONSERVER_DataTypeByteSize(to_triton(out_info.type));
+    OBufferDescr buffer;
+    buffer.size = volume(output_shape.begin(), output_shape.end()) * t_size;
```
mzient, line 81: Doesn't that work?
```suggestion
    buffer.size = volume(output_shape) * t_size;
```

banasraf: it works

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -0,0 +1,99 @@
...
+    inputs = [self._get_input(batch, name) for batch, name in zip(batches, self.input_names)]
+    outputs = [t_client.InferRequestedOutput(name) for name in self.output_names]
+    results = self.client.infer(model_name=self.model_name, inputs=inputs, outputs=outputs)
+    return [results.as_numpy(name) for name in self.output_names]
+
+  def run_tests(self, data, compare_to, n_infers=-1, eps=1e-7):
+    generator = data if n_infers < 1 else islice(cycle(data), n_infers)
+    it = -1
+    for batches in generator:
+      it += 1
```
JanuszL, line 69:
```suggestion
    for it, batches in enumerate(generator):
```

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py
```diff
@@ -0,0 +1,43 @@
...
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import nvidia.dali as dali
+import nvidia.dali.fn as fn
+import multiprocessing as mp
+import argparse
+
+@dali.pipeline_def(batch_size=1, num_threads=mp.cpu_count(), device_id=0)
```
JanuszL, line 27: This is a lot of threads. How about `max(mp.cpu_count(), 4)`?

qa/L0_DALI_GPU_ensemble/setup.sh
```diff
@@ -0,0 +1,33 @@
...
+mkdir -p model_repository/dali_1/1 
+python model_repository/dali_1/pipeline.py model_repository/dali_1/1/model.dali 
+echo "dali_1 ready"
+
+mkdir -p model_repository/dali_2/1
+python model_repository/dali_2/pipeline.py model_repository/dali_2/1/model.dali 
+echo "dali_2 ready"
+
+mkdir -p model_repository/dali_ensemble/1
+echo "dali_ensemble ready"
```
JanuszL, line 33: Newline

qa/L0_DALI_GPU_ensemble/model_repository/dali_2/pipeline.py
```diff
@@ -0,0 +1,43 @@
+# The MIT License (MIT)
```
JanuszL, line 1: Can you unify with `qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py`. You can add a switch selecting which pipeline is used.

qa/L0_DALI_GPU_ensemble/test.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+# FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+echo "RUN CLIENT"
+python client.py
```
JanuszL, line 25: Newline

qa/L0_DALI_GPU_ensemble/test.sh
```diff
@@ -0,0 +1,25 @@
...
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+# FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+echo "RUN CLIENT"
```
JanuszL, line 24: Maybe you can merge setup.sh with test.sh? Or you prefer to have a clear separation between test and setup for it?

src/dali_backend.cc
```diff
@@ -178,7 +200,25 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
     dali_executor_ = std::make_unique<DaliExecutor>(std::move(pipeline));
   }
 
-
+  std::vector<IDescr> GenerateInputs(TritonRequest& request) {
+    uint32_t input_cnt = request.InputCount();
+    std::vector<IDescr> ret;
```
JanuszL, line 205:
```suggestion
    std::vector<IDescr> ret(input_cnt);
```

src/dali_backend.cc
```diff
@@ -178,7 +200,25 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+    for (uint32_t input_idx = 0; input_idx < input_cnt; ++input_idx) {
+      auto input = request.InputByIdx(input_idx);
+      auto input_byte_size = input.ByteSize();
+      auto input_buffer_count = input.BufferCount();
+      std::vector<IBufferDescr> buffers;
```
JanuszL, line 210:
```suggestion
      std::vector<IBufferDescr> buffers(input_buffer_count);
```

src/dali_backend.cc
```diff
@@ -178,7 +200,25 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+      for (uint32_t buffer_idx = 0; buffer_idx < input_buffer_count; ++buffer_idx) {
+        auto buffer = input.GetBuffer(buffer_idx);
+        ENFORCE(buffer.device == device_type_t::CPU || buffer.device_id == device_id_,
+                "GPU input must reside on the same device that the model instance.");
+        buffers.push_back(buffer);
+      }
+      ret.push_back({input.Meta(), buffers});
```
JanuszL, line 217: Would this make any difference?
```suggestion
      ret.push_back({input.Meta(), std::move(buffers)});
```

src/dali_backend.cc
```diff
@@ -484,10 +498,6 @@ TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(TRITONBACKEND_ModelInstan
             responses[i], TRITONSERVER_ResponseCompleteFlag::TRITONSERVER_RESPONSE_COMPLETE_FINAL,
             error),
         make_string("Failed sending response, idx ", i));
-    LOG_IF_ERROR(
```
JanuszL, line 487: Why removed?

src/dali_executor/dali_executor.cc
```diff
@@ -21,28 +21,56 @@
...
+  std::vector<IDescr> c_inputs{};
   for (auto& inp : inputs) {
-    assert(inp.shape.num_elements() * dali_type_size(inp.type) <= inp.buffer.size());
-    pipeline_.SetInput(inp.buffer.data(), inp.name.c_str(), inp.device, inp.type, inp.shape);
+    size_t inp_size = inp.meta.shape.num_elements() * dali_type_size(inp.meta.type);
+    if (inp.buffers.size() == 1) {
+      assert(inp_size <= inp.buffers[0].size);
+      c_inputs.push_back(inp);
+    } else {
+      c_inputs.push_back(ScheduleInputCopy(inp));
```
JanuszL, line 42: I would add a comment that you are merging requests into one continuous buffer.

src/dali_executor/dali_executor.cc
```diff
@@ -21,28 +21,56 @@
...
+      assert(inp_size <= c_inputs.back().buffers[0].size);
+    }
+  }
+  thread_pool_.RunAll();
```
JanuszL, line 46: It was not clear for me that ScheduleInputCopy used the thread_pool_ under the hood. Maybe there should be a separate method, like `ExecuteInputCopu` that will pair with ScheduleInputCopy and will abstract away thread_pool_ usage.

src/dali_executor/dali_executor.cc
```diff
@@ -21,28 +21,56 @@
...
+    buffer = &cpu_buffers_[input.meta.name];
+  } else {
+    buffer = &gpu_buffers_[input.meta.name];
+  }
+  size_t size = 0;
+  for (auto& buf : input.buffers)
+    size += buf.size;
+  buffer->Allocate(size);
+  for (auto& buf : input.buffers) {
+    auto origin = buffer->Reserve(buf.size);
```
JanuszL, line 65:
```suggestion
    auto target= buffer->Reserve(buf.size);
```
As I understand you copy to it, not from it.

src/dali_executor/io_buffer.cc
```diff
@@ -0,0 +1,42 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#include "src/dali_executor/io_buffer.h"
+
+namespace triton { namespace backend { namespace dali {
+
+void CopyMem(device_type_t dst_dev, void *dst, device_type_t src_dev, const void *src, size_t size,
```
JanuszL, line 27: Maye this could be a set of 4 template specializations?

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,135 @@
...
+#ifndef TRITONDALIBACKEND_IO_BUFFER_H
+#define TRITONDALIBACKEND_IO_BUFFER_H
+
+#include "src/dali_executor/io_descriptor.h"
+#include "src/dali_executor/utils/dali.h"
+
+namespace triton { namespace backend { namespace dali {
+
+
+void CopyMem(device_type_t dst_dev, void *dst, device_type_t src_dev, const void *src, size_t size,
```
JanuszL, line 32:
```suggestion
void MemCopy(device_type_t dst_dev, void *dst, device_type_t src_dev, const void *src, size_t size,
```

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Cancel all reservations. No memory is deallocated.
+   */
+  virtual void Clear() = 0;
+
+  /**
+   * Allocate `size` bytes of memory.
+   * If the buffer's capacity is greater or equal to size, this function is a no-op.
+   */
+  virtual void Allocate(size_t size) = 0;
```
JanuszL, line 52: I would swap names Allocate and Reserve. As Reserve usually doesn't provide any useful memory, while allocate does. In this case it is the opposite.

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,135 @@
...
+
+  virtual size_t Capacity() const = 0;
```
JanuszL, line 54: Can you add docs for the remaining functions?

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,135 @@
...
+
+  size_t Capacity() const override {
+    return buffer_.size();
+  }
+
+  device_type_t DeviceType() const override {
+    return Dev;
+  }
+
+  IBufferDescr GetDescr() const override {
```
JanuszL, line 109: It confuses me that the return type depend on the consents of the object. Seems to be error prone.

src/utils/triton.h
```diff
@@ -0,0 +1,170 @@
...
+
+namespace triton { namespace backend { namespace dali {
+
+
+/**
+ * Converts TRITONSERVER_DataType to dali_data_type_t
+ */
+inline dali_data_type_t to_dali(TRITONSERVER_DataType t) {
+  assert(t >= 0 && t <= 13);
+  // Use the trick, that TRITONSERVER_DataType and dali_data_type_t
```
JanuszL, line 39: What if TRITON or DALI changes it?

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,135 @@
...
+    filled_ += size;
+    return origin;
+  }
+
+  void Clear() override {
+    filled_ = 0;
+  }
+
+  void Allocate(size_t size) override {
+    if (size > buffer_.size())
```
JanuszL, line 97: Maybe it should not be possible to allocate when filled_ != 0?

src/dali_executor/io_buffer.test.cc
```diff
@@ -0,0 +1,84 @@
...
+
+namespace triton { namespace backend { namespace dali { namespace test {
+
+template<device_type_t Dev>
+void test_buffer(IOBuffer<Dev> &buffer) {
+  const uint8_t N = 10;
+  const size_t size = N * (N + 1) / 2;
+  buffer.Clear();
+  buffer.Allocate(size);
+  REQUIRE(buffer.Capacity() >= size);
```
JanuszL, line 36:
```suggestion
  REQUIRE(buffer.Capacity() == size);
```
Or you want to implement grow factor in the future?

src/dali_executor/io_buffer.test.cc
```diff
@@ -0,0 +1,84 @@
...
+  for (uint8_t i = 1; i <= N; ++i) {
+    std::vector<uint8_t> chunk(i, i);
+    auto origin = buffer.Reserve(i);
```
JanuszL, line 39:
```suggestion
    auto dst = buffer.Reserve(i);
```

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py
mzient: ...did you mean `min`?

JanuszL: Right, min.

src/dali_backend.cc
banasraf: Because TritonRequest class handles releasing the request in RAII fashion

src/dali_executor/io_buffer.h
banasraf: It's like with std::vector data() method. It returns `T*` for non-const vec and `const T*` for const vec

qa/L0_DALI_GPU_ensemble/model_repository/dali_2/pipeline.py
banasraf: You mean merging it into one file? What is a benefit of doing this? 
I put those file in the model directories thinking that in the future that might simplify automating testing against DALI pipelines (e.g. automatically importing pipelines for each model and running them)

qa/L0_DALI_GPU_ensemble/test.sh
banasraf: It's how CI works. There's a separate setup and test stage

qa/L0_DALI_GPU_ensemble/client.py
banasraf: done

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -0,0 +1,99 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+# FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import nvidia.dali as dali
+import multiprocessing as mp
```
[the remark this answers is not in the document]
banasraf: removed

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py
banasraf: done

src/dali_executor/dali_executor.cc
banasraf: done

src/dali_executor/dali_executor.cc
banasraf: renamed to dst

src/dali_executor/io_buffer.cc
banasraf: device type is comes from runtime, so if I made this a static parameter to a template, I would have to then do a switch case to call it.

src/dali_executor/io_buffer.h
banasraf: done

src/dali_executor/io_buffer.h
banasraf: done

src/dali_executor/io_buffer.h
banasraf: done

src/dali_executor/io_buffer.h
banasraf: true. done

src/dali_executor/io_buffer.test.cc
banasraf: done

src/dali_executor/io_buffer.test.cc
banasraf: done

src/dali_executor/dali_executor.cc
banasraf: done

src/utils/triton.h
banasraf: I'll change it next time I'll be on this file (it will happen soon), to not inflate this PR further

src/dali_backend.cc
banasraf: It might. Done

src/dali_backend.cc
banasraf: I used .reserve()

src/dali_backend.cc
banasraf: I used .reserve()

banasraf:
I resolved a bug by adding `DeviceGuard` to the request processing function and adding `cudaStreamSynchronize` after output copy from DALI pipeline.

qa/L0_DALI_GPU_ensemble/model_repository/dali_2/pipeline.py
JanuszL: To reduce the repetition, but just a suggestion.

qa/L0_DALI_GPU_ensemble/test.sh
JanuszL: OK

src/utils/triton.h
JanuszL: OK

src/dali_executor/dali_executor.cc
```diff
@@ -21,28 +21,67 @@
...
+  buffer->Clear();
+  size_t size = 0;
+  for (auto& buf : input.buffers)
+    size += buf.size;
+  buffer->Reserve(size);
+  for (auto& buf : input.buffers) {
+    auto dst = buffer->Allocate(buf.size);
+    thread_pool_.AddWork([buffer, dst, buf](int) {
+      MemCopy(buffer->DeviceType(), dst, buf.device, buf.data, buf.size);
+    });
```
JanuszL, line 70:
```suggestion
    },0, true);
```
I don't know how fast Allocate is but I think it should be fine to start processing immediately.

src/dali_executor/dali_executor.h
```diff
@@ -23,40 +23,64 @@
...
-  void SetupInputs(const std::vector<IODescr<owns>>& inputs);
+  void SetupInputs(const std::vector<IDescr>& inputs);
+
+  /**
+   * Schedule copy to a continous buffer and return IDecr to the new buffer.
+   */
+  IDescr ScheduleInputCopy(const IDescr& buffers);
+
+  /**
+   * Run previously scheduled copy of input buffers.
```
JanuszL, line 66:
```suggestion
   * Run copies scheduled by ScheduleInputCopy.
```

src/dali_executor/dali_pipeline.cc
```diff
@@ -70,11 +70,20 @@ void DaliPipeline::SetInput(const void* ptr, const char* name, device_type_t sou
...
+
+
 void DaliPipeline::PutOutput(void* destination, int output_idx, device_type_t destination_device) {
   assert(destination != nullptr);
   assert(output_idx >= 0);
-  daliOutputCopy(&handle_, destination, output_idx, destination_device,
-                 destination_device == CPU ? nullptr : output_stream_, 0);
+  DeviceGuard dg(device_id_);
+  daliOutputCopy(&handle_, destination, output_idx, destination_device, output_stream_, 0);
+  CUDA_CALL(cudaStreamSynchronize(output_stream_));
```
JanuszL, line 86: DO we want to sync after each copy? Maybe we can do it once after all copies have been scheduled?

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,149 @@
...
+    descr.size = filled_;
+    descr.device = Dev;
+    descr.device_id = device_id;
+    return descr;
+  }
+
+ private:
+  buffer_t<uint8_t, Dev> buffer_;
+  size_t filled_ = 0;
+  int device_id = 0;
```
JanuszL, line 144:
```suggestion
  int device_id_ = 0;
```

src/utils/triton.h
```diff
@@ -0,0 +1,177 @@
...
+
+  uint32_t BufferCount() const {
+    return buffer_cnt_;
+  }
+
+  /**
+   * Request an input buffer.
+   * idx  number index
+   * device_type_t preferred device type
+   * device_id preferred device id
```
JanuszL, line 128: Don't you want to use doxygen convention?

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -0,0 +1,107 @@
...
+      assert(len(results) == len(ref))
+      for out_i, (out, ref_out) in enumerate(zip(results, ref)):
+        assert out.shape == ref_out.shape
+        if not np.allclose(out, ref_out, atol=eps):
+          print("Test failure in iteration", it)
+          print("Output", out_i)
+          print("Expected:\n", ref_out)
+          print("Actual:\n", out)
+          return
+      print('PASS iteration:', it)
```
JanuszL, line 78: Maybe it should be a function that you may use in other tests?

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+# TODO: Use actual DALI pipelines to calculate ground truth
```
mzient, line 81: Nitpick: don't do that _instead_ of this test - sure, using a DALI pipeline would allow you to check for bit-exact results, but if there's some interaction between this process and DALI that makes DALI produce consistently wrong results (think: all zeros?) you wouldn't know.

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -0,0 +1,107 @@
...
+      ref = compare_to(*batches)
+      assert(len(results) == len(ref))
+      for out_i, (out, ref_out) in enumerate(zip(results, ref)):
+        assert out.shape == ref_out.shape
+        if not np.allclose(out, ref_out, atol=eps):
+          print("Test failure in iteration", it)
+          print("Output", out_i)
+          print("Expected:\n", ref_out)
+          print("Actual:\n", out)
+          return
```
mzient, line 77: Wouldn't this just print the differences and quit? I think you should indicate the failure somehow.
```suggestion
          assert False
```

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -0,0 +1,107 @@
...
+    outputs = [t_client.InferRequestedOutput(name) for name in self.output_names]
+    results = self.client.infer(model_name=self.model_name, inputs=inputs, outputs=outputs)
+    return [results.as_numpy(name) for name in self.output_names]
+
+  def run_tests(self, data, compare_to, n_infers=-1, eps=1e-7):
+    generator = data if n_infers < 1 else islice(cycle(data), n_infers)
+    for it, batches in enumerate(generator):
+      results = self.run_inference(batches)
+      ref = compare_to(*batches)
+      assert(len(results) == len(ref))
```
mzient, line 69: Hmm... if this condition fails, you'll never examine the results in detail.

qa/L0_DALI_GPU_ensemble/client.py
banasraf: What do you mean? The purpose of this test is not to test DALI, but to test DALI backend, so I think it's a reasonable to assume that the same pipelines run locally and inside Triton server should give the same results.

qa/L0_DALI_GPU_ensemble/client.py
banasraf: I plan to refactor all other tests to use these functions in a future PR. For now I didn't want to create another file for this.

qa/L0_DALI_GPU_ensemble/client.py
banasraf: fixed

src/utils/triton.h
banasraf: Fixed. (is it doxygen convention now?)

src/dali_executor/io_buffer.h
banasraf: done

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,149 @@
...
+      CUDA_CALL_GUARD(cudaGetDevice(&device_id));
+    }
+  }
+
+  uint8_t *Allocate(size_t size) override {
+    ENFORCE(filled_ + size <= buffer_.size(),
+            make_string("Not enough memory reserved (", Capacity(),
+                        " bytes) to allocate "
+                        "a chunk of size ",
+                        size));
```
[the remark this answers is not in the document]
banasraf: fixed

src/dali_executor/dali_pipeline.cc
banasraf: Moved synchronization to another function that is called after copying all of the outputs.

src/dali_executor/dali_executor.h
banasraf: done

src/dali_executor/dali_executor.cc
banasraf: done

src/utils/triton.h
```diff
@@ -0,0 +1,180 @@
...
+      return TRITONSERVER_MEMORY_GPU;
+    default:
+      throw std::invalid_argument("Unknown memory type");
+  }
+}
+
+class TritonInput {
+ public:
+  TritonInput(TRITONBACKEND_Input *handle) : handle_(handle) {
+    TRITONSERVER_DataType input_datatype;
```
szalpal, line 96: If we go the class way (I like it btw), maybe `TRITONBACKEND_Input` should be enclosed inside `TritonInput`? Why would we need it outside?

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py
```diff
@@ -0,0 +1,43 @@
...
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import nvidia.dali as dali
+import nvidia.dali.fn as fn
+import multiprocessing as mp
+import argparse
+
+@dali.pipeline_def(batch_size=1, num_threads=min(mp.cpu_count(), 4), device_id=0)
```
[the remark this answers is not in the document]
banasraf: Why? This batch size isn't relevant for the DALI backend, right?

qa/L0_DALI_GPU_ensemble/client.py
mzient: OK, I forgot that's in another process. I take it back.

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py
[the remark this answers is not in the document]
mzient: It sort of is - indeed, it would be best to check that it works with _variable_ batch size, since by far Triron is the only client of DALI that uses variable batch size though C API.

cmake/dali.cmake
```diff
@@ -44,5 +44,5 @@ function(get_dali_paths DALI_INCLUDE_DIR_VAR DALI_LIB_DIR_VAR DALI_LIBRARIES_VAR
 
     set(${DALI_INCLUDE_DIR_VAR} ${DALI_INCLUDE_DIR} PARENT_SCOPE)
     set(${DALI_LIB_DIR_VAR} ${DALI_LIB_DIR} PARENT_SCOPE)
-    set(${DALI_LIBRARIES_VAR} dali dali_kernels dali_operators PARENT_SCOPE)
+    set(${DALI_LIBRARIES_VAR} dali dali_core dali_kernels dali_operators PARENT_SCOPE)
```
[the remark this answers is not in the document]
banasraf: It didn't work without it (undefined symbols).

src/dali_executor/utils/dali.h
```diff
@@ -53,6 +68,25 @@ inline int64_t dali_type_size(dali_data_type_t type) {
...
+  for (auto &shape : shapes)
+    num_samples += shape.num_samples();
+  TensorListShape<Dims> result(num_samples, ndims);
+  int64_t ti = 0;
+  for (auto &shape : shapes) {
+    for (int64_t j = 0; j < shape.num_samples(); ++j) {
+      result.set_tensor_shape(ti, shape.tensor_shape_span(j));
+      ++ti;
+    }
+  }
```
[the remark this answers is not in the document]
banasraf: but `.set_tensor_shape` is for setting  tensor shape and `shape` here is a tensor list shape.

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py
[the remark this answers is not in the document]
banasraf: Does it really matter? This batch size is discarded anyway.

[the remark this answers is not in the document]
banasraf: AThis `batch_size`  parameter is discarded when we deserialize pipeline and it's assigned the `max_batch_size` parameter from the config. In the tests, the batch size is variable.

qa/L0_DALI_GPU_ensemble/model_repository/dali_ensemble/config.pbtxt
```diff
@@ -0,0 +1,96 @@
...
+  }
+]
+ensemble_scheduling {
+  step [
+    {
+      model_name: "dali_1"
+      model_version: -1
+      input_map {
+        key: "INPUT_0"
+        value: "INPUT0"
```
mzient, line 60: Nitpick: These names make it hard to read - when they're next to one another, it's easy to notice the underscore, not so much otherwise.

src/utils/triton.h
banasraf: I think creating the `TritonInput` object is a responsibility of `TritonRequest` class.

src/dali_executor/utils/dali.h
```diff
@@ -53,6 +68,25 @@ inline int64_t dali_type_size(dali_data_type_t type) {
...
 
+template<typename Container, int Dims = -1>
+TensorListShape<Dims> cat_list_shapes(const Container &shapes) {
+  if (shapes.empty())
+    return TensorListShape<Dims>(0);
+  int64_t num_samples = 0;
+  int64_t ndims = shapes.begin()->sample_dim();
+  for (auto &shape : shapes)
+    num_samples += shape.num_samples();
+  TensorListShape<Dims> result(num_samples, ndims);
```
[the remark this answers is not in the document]
banasraf: Added some enforces

src/dali_executor/test/test_utils.h
```diff
@@ -62,16 +62,26 @@ constexpr dali_data_type_t dali_data_type() {
...
-  buffer.clear();
-  std::generate_n(std::back_inserter(buffer), shape.num_elements(), generator);
-  IODescr<false> dscr;
-  dscr.name = name;
-  dscr.shape = shape;
-  dscr.type = dali_data_type<T>();
-  dscr.buffer = span<char>(reinterpret_cast<char*>(buffer.data()), sizeof(T) * buffer.size());
-  dscr.device = device_type_t::CPU;
+IDescr RandomInput(std::vector<std::vector<T>>& buffers, const std::string& name,
+                   const std::vector<TensorListShape<>>& shapes, const R& generator) {
```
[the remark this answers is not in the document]
banasraf: added

qa/L0_DALI_GPU_ensemble/model_repository/dali_ensemble/config.pbtxt
banasraf: renamed to DALI_INPUT/OUTPUT_0/1 for DALI ins/outs and INPUT_0/1 for ensemble ins/outs.

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/pipeline.py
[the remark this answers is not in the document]
mzient: OK.

src/utils/triton.h
```diff
@@ -0,0 +1,180 @@
...
+ public:
+  DALI_INHERIT_UNIQUE_HANDLE(TRITONBACKEND_Request *, TritonRequest)
+
+  static void DestroyHandle(TRITONBACKEND_Request *request) {
+    LOG_IF_ERROR(TRITONBACKEND_RequestRelease(
+                     request, TRITONSERVER_RequestReleaseFlag::TRITONSERVER_REQUEST_RELEASE_ALL),
+                 make_string("Failed releasing a request."));
+  }
+
+  uint32_t InputCount() {
```
mzient, line 165:
```suggestion
  uint32_t InputCount() const {
```

src/utils/triton.h
```diff
@@ -0,0 +1,180 @@
...
+    uint32_t input_cnt;
+    TRITON_CALL_GUARD(TRITONBACKEND_RequestInputCount(handle_, &input_cnt));
+    return input_cnt;
+  }
+
+  TritonInput InputByIdx(uint32_t idx) {
```
mzient, line 171:
```suggestion
  TritonInput InputByIdx(uint32_t idx) const {
```
Even if the call modifies the request, it does not modify the handle/wrapper.

mzient: A general remark about this class:
I think that adding C++ wrappers around Triton handles is a great idea, but they should be complete and that should be a separate PR. I will turn a blind eye on this here, but that mandates a follow-up IMO.
Also, there probably should be a TritonRequestView non-owning variant. We can discuss the pattern if you wish.

src/dali_executor/io_descriptor.h
```diff
@@ -28,169 +28,37 @@
...
-  std::string name;
-  dali_data_type_t type;
+struct BufferDescr {
   device_type_t device;
   int device_id;
-  TensorListShape<> shape;
-  span<T> buffer;
-};
+  T* data;
+  size_t size;
```
mzient, line 36: Does this type need to be a POD? If not, I'd strongly recommend adding field initializers. Leaving the fields uninitialized can lead to failures when a new field is added. We've seen this exact kind of failure in DALI before, so it's not theoretical.

src/dali_executor/io_descriptor.h
```diff
@@ -28,169 +28,37 @@
...
-    detail::allocate<StorageBackend>(&owned_mem, owned_mem_capacity);
-  }
-
-
-  void dealloc() {
-    detail::deallocate<StorageBackend>(owned_mem);
-  }
+struct IOMeta {
+  std::string name;
+  dali_data_type_t type;
```
mzient, line 50:
```suggestion
  dali_data_type_t type = {};
```

src/dali_executor/test/test_utils.h
```diff
@@ -61,17 +61,39 @@ constexpr dali_data_type_t dali_data_type() {
...
+    buf_dscr.data = buffer.data();
+    buf_dscr.size = sizeof(T) * buffer.size();
+    buf_dscr.device = device_type_t::CPU;
+    buf_descs.push_back(buf_dscr);
+  }
+  IDescr dscr;
+  dscr.meta.name = name;
+  dscr.meta.shape = cat_list_shapes(shapes);
+  dscr.meta.type = dali_data_type<T>();
+  dscr.buffers = buf_descs;
```
mzient, line 96: This is dangerous, because `IDescr::IDescr()` doesn't initialize all fields. If `IDescr` is a POD, use `IDescr srcr = {};`. If `IDescr` is not a POD, then it should have field initializers (in-class field initializers are highly preferable to initializer lists because these, again, can miss some fields _and_ would need to be duplicated in all constructors).

src/utils/triton.h
```diff
@@ -0,0 +1,180 @@
...
+    size_t size;
+    TRITONSERVER_MemoryType mem_type = to_triton(device);
+    int64_t mem_type_id = device_id;
+    TRITON_CALL_GUARD(
+        TRITONBACKEND_InputBuffer(handle_, idx, &data, &size, &mem_type, &mem_type_id));
+    IBufferDescr descr;
+    descr.device = to_dali(mem_type);
+    descr.device_id = mem_type_id;
+    descr.data = data;
+    descr.size = size;
```
mzient, line 141: Add initializers to `BufferDescr` - otherwise this code is not future proof and a potential correctness or even security issue.

src/dali_executor/executor.test.cc
```diff
@@ -30,70 +30,88 @@
...
+    size_t inp_size = 0;
+    for (auto &inp_buffer : input_buffers)
+      inp_size += inp_buffer.size();
+    std::vector<float> output_buffer(inp_size);
+    std::vector<ODescr> output_vec(1);
+    auto &outdesc = output_vec[0];
+    OBufferDescr buf_descr;
+    buf_descr.device = device_type_t::CPU;
+    buf_descr.data = output_buffer.data();
+    buf_descr.size = output_buffer.size() * sizeof(decltype(output_buffer)::size_type);
```
mzient, line 62: Make sure that `BufferDescr` initializes its fields.

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,155 @@
...
+  device_type_t DeviceType() const override {
+    return Dev;
+  }
+
+  IBufferDescr GetDescr() const override {
+    IBufferDescr descr;
+    descr.data = buffer_.data();
+    descr.size = filled_;
+    descr.device = Dev;
+    descr.device_id = device_id_;
```
mzient, line 134: Dangerous due to potentially uninitialized fields. Fix the constructor of `BuferDescr` or use `descr = {};`

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,155 @@
...
+    return descr;
+  }
+
+  OBufferDescr GetDescr() override {
+    OBufferDescr descr;
+    descr.data = buffer_.data();
+    descr.size = filled_;
+    descr.device = Dev;
+    descr.device_id = device_id_;
```
mzient, line 143: Dangerous due to potentially uninitialized fields. Fix the constructor of `BuferDescr` or use `descr = {};`

src/dali_executor/test/test_utils.h
```diff
@@ -61,17 +61,39 @@ constexpr dali_data_type_t dali_data_type() {
...
+IDescr RandomInput(std::vector<std::vector<T>>& buffers, const std::string& name,
+                   const std::vector<TensorListShape<>>& shapes, const R& generator) {
+  std::vector<IBufferDescr> buf_descs;
+  assert(buffers.size() == shapes.size());
+  for (size_t i = 0; i < buffers.size(); ++i) {
+    auto& buffer = buffers[i];
+    auto& shape = shapes[i];
+    buffer.clear();
+    std::generate_n(std::back_inserter(buffer), shape.num_elements(), generator);
+    IBufferDescr buf_dscr;
```
mzient, line 86: Dangerous due to potentially uninitialized fields. Fix the constructor of `BuferDescr` or use `descr = {};`

src/utils/triton.h
mzient: This is how you can have an owning and non-owning variant and implement the wrapper functions only once:
```C++
template <class Actual>
class TritonRequestWrapper {
 public:
  uint32_t InputCount() const {
    uint32_t input_cnt;
    TRITON_CALL_GUARD(TRITONBACKEND_RequestInputCount(This().handle(), &input_cnt));
    return input_cnt;
  }

  TritonInput InputByIdx(uint32_t idx) const {
    TRITONBACKEND_Input *input;
    TRITON_CALL_GUARD(TRITONBACKEND_RequestInputByIndex(This().handle(), idx, &input));
    return TritonInput(input);
  }

  // add whatever other functions you need

 private:
  Actual &This() noexcept {
    return static_cast<Actual &>(*this);
  }

  const Actual &This() const noexcept {
    return static_cast<const Actual &>(*this);
  }
};


class TritonRequest : public UniqueHandle<TRITONBACKEND_Request *, TritonRequest>,
                      public TritonRequestWrapper<TritonRequest> {
 public:
  DALI_INHERIT_UNIQUE_HANDLE(TRITONBACKEND_Request *, TritonRequest)

  static void DestroyHandle(TRITONBACKEND_Request *request) {
    LOG_IF_ERROR(TRITONBACKEND_RequestRelease(
                     request, TRITONSERVER_RequestReleaseFlag::TRITONSERVER_REQUEST_RELEASE_ALL),
                 make_string("Failed releasing a request."));
  }
};

class TritonRequestView : public TritonRequestWrapper<TritonRequestView> {
 public:
  TritonRequestView() = default;

  TritonRequestView(TRITONBACKEND_Request *req) : handle_(req) {}  // NOLINT

  TRITONBACKEND_Request *handle() const noexcept {
    return handle_;
  }
 private:
  TRITONBACKEND_Request *handle_ = nullptr;
};
```

banasraf: I totally agree with the incompleteness of the TritonRequest, but I wanted to implement only the absolute minimum to not inflate this PR. Before I start adding new features, in the next PR I will implement all of the Triton types properly in a separate PR.

banasraf: That wrapper is a neat trick. Thanks!

src/utils/triton.h
banasraf: done

src/dali_executor/io_descriptor.h
banasraf: fixed BufferDescr constructor

src/dali_executor/io_descriptor.h
banasraf: added initialization

src/dali_backend.h
```diff
@@ -144,42 +48,42 @@ std::vector<IODescr<true>> GenerateInputs(TRITONBACKEND_Request* request) {
  * processing.
  * @return A descriptor that wraps Triton output
  */
-std::vector<IODescr<false>> AllocateOutputs(
-    TRITONBACKEND_Request* request, TRITONBACKEND_Response* response,
-    const std::vector<shape_and_type_t>& shapes_and_types,
-    const std::unordered_map<std::string, int>& output_order) {
+std::vector<ODescr> AllocateOutputs(TRITONBACKEND_Request* request,
```
szalpal, line 51: Why not `TritonRequest` here?

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,155 @@
...
+class IOBuffer : public IOBufferI {
+ public:
+  IOBuffer(size_t size = 0) : buffer_() {
+    buffer_.resize(size);
+    if (Dev == device_type_t::GPU) {
+      CUDA_CALL_GUARD(cudaGetDevice(&device_id_));
+    }
+  }
+
+  uint8_t *Allocate(size_t size) override {
```
mzient, line 101: 1. The name of this function is unfortunate to say the least. Without reading the code `Allocate` and `Reserve` seem like synonyms.
2. There's no alignment support. x86 or ARM don't require aligned memory to work _at all_, but they do require natural alignment to perform well. I think that the GPU requires natural alignment at least for fundamental types, though. Most likely this worked by sheer luck.

```suggestion
  uint8_t *GetChunk(size_t size, size_t alignment = alignof(std::max_align_t)) override {
```

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,155 @@
...
+    ENFORCE(filled_ + size <= buffer_.size(),
+            make_string("Not enough memory reserved (", Capacity(), " bytes) to allocate ",
+                        "a chunk of size ", size));
+    auto origin = buffer_.data() + filled_;
+    filled_ += size;
+    return origin;
```
mzient, line 107:
```suggestion
    uint8_t *base = buffer_.data();
    uint8_t *chunk_start = dali::mm::align_ptr(base + filled_, alignment);
    size_t offset = chunk_start - base;
    ENFORCE(offset + size <= buffer_.size(),
            make_string("Not enough memory reserved to allocate a chunk of size ", size,
            " with required alignment ", alignment, "."
            "\n  Capacity: ", capacity,
            "\n  Occupied: ", filled_,
            "\n  Occupied (incl. padding): ", offset,
            "\n  Requested size: ", size));
    filled_ = offset + size;
    return base;
```

src/dali_backend.h
banasraf: I will move the whole response and outputs logic to the C++ API in the next PR

mzient (approved):

src/dali_backend.h
mzient: If anything, it should be the view (non-owning).

src/dali_executor/io_buffer.h
```diff
@@ -0,0 +1,155 @@
...
+  IOBufferI() {}
+};
+
+template<typename T, device_type_t Dev>
+using buffer_t = std::conditional_t<Dev == device_type_t::CPU, std::vector<T>, DeviceBuffer<T>>;
+
+template<device_type_t Dev>
+class IOBuffer : public IOBufferI {
+ public:
+  IOBuffer(size_t size = 0) : buffer_() {
```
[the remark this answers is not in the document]
banasraf: done

banasraf:
Follow-ups: 
* Implement C++ wrappers for the rest of the Triton API
* Move the `TritonClient` class from the python test to a separate module and use it in the rest of the tests.

src/dali_executor/dali_pipeline.h
```diff
@@ -116,13 +122,30 @@ class DaliPipeline {
...
   void Reset() {
     ReleasePipeline();
     CreatePipeline();
   }
 
+  int DeviceId() {
+    return device_id_;
+  }
+
+  int NumThreadsArg() {
```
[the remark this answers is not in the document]
banasraf: It's because it's not the actual number of threads used by the pipeline but the argument passed to the constructor (it might be equal -1). I have no access to the actual number of threads used by the pipeline.

src/dali_executor/utils/utils.h
```diff
@@ -54,6 +54,29 @@ std::vector<int64_t> array_shape(TensorListShape<ndims> tls) {
...
+  ENFORCE(Dims == -1 || Dims == ndims,
+          make_string("Cannot convert shape with ", ndims, " dimensions to shape with ", Dims,
+                      " dimensions."));
+  TensorListShape<Dims> result(num_samples, ndims);
+  int64_t ti = 0;
+  for (auto &shape : shapes) {
+    ENFORCE(shape.sample_dim() == ndims, "Cannot concatenate shapes of different dimensionality");
+    for (int64_t j = 0; j < shape.num_samples(); ++j) {
+      result.set_tensor_shape(ti, shape.tensor_shape_span(j));
+      ++ti;
```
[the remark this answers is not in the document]
banasraf: done

src/dali_executor/test/test_utils.h
```diff
@@ -61,17 +61,39 @@ constexpr dali_data_type_t dali_data_type() {
...
-  dscr.name = name;
-  dscr.shape = shape;
-  dscr.type = dali_data_type<T>();
-  dscr.buffer = span<char>(reinterpret_cast<char*>(buffer.data()), sizeof(T) * buffer.size());
-  dscr.device = device_type_t::CPU;
+IDescr RandomInput(std::vector<std::vector<T>>& buffers, const std::string& name,
+                   const std::vector<TensorListShape<>>& shapes, const R& generator) {
+  std::vector<IBufferDescr> buf_descs;
+  assert(buffers.size() == shapes.size());
+  for (size_t i = 0; i < buffers.size(); ++i) {
```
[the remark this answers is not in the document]
banasraf: I added `buffers = std::vector<std::vector<T>>(shapes.size());`

szalpal (approved):

JanuszL (approved):

szalpal (approved):

PR #59, opened 2021-05-11
Add RN50-TRT example to the CI

szalpal:
Relates to !39

Adding this example needs slight refactoring, thus it's done separately. The code of the example is not new - it has been there before (refer - `dali_backend/docs/examples/resnet50-trt`). The only changes are in the `setup.sh` file (change in the way of setting the example up) and `serialize_dali_pipeline.py` (adjusting for DALI 1.0 decorator API).


klecki (commented):
If this is a copy and paste from the example I would suggest looking into a way of running those examples in CI rather than duplicating the code that will go out of sync at some point.

szalpal:
> If this is a copy and paste from the example I would suggest looking into a way of running those examples in CI rather than duplicating the code that will go out of sync at some point.

This test is going to be the reference code for the example, due to many reasons: `qa/L0_...` pattern stipulated by the CI and peculiar `setup.sh` and `test.sh` scripts (which probably shouldn't be included in the example) to name two. Thus, when it's merged, I'll think about how to cleanly express `qa` tests as the examples. In this PR I won't remove the example code, not to clutter the PR more.

qa/L0_resnet50_trt/setup.sh
```diff
@@ -0,0 +1,15 @@
...
+mkdir -p ensemble_dali_resnet50/1
+mkdir -p resnet50_trt/1
+
+python onnx_exporter.py --save model.onnx
+trtexec --onnx=model.onnx --saveEngine=resnet50_trt/1/model.plan --explicitBatch --minShapes=input:1x3x224x224 --optShapes=input:1x3x224x224 --maxShapes=input:256x3x224x224 --fp16
+python serialize_dali_pipeline.py --save ./dali/1/model.dali
+
+echo "Resnet50 model ready."
+
+popd
```
banasraf, line 15: newline

banasraf:
> > If this is a copy and paste from the example I would suggest looking into a way of running those examples in CI rather than duplicating the code that will go out of sync at some point.
> 
> This test is going to be the reference code for the example, due to many reasons: `qa/L0_...` pattern stipulated by the CI and peculiar `setup.sh` and `test.sh` scripts (which probably shouldn't be included in the example) to name two. Thus, when it's merged, I'll think about how to cleanly express `qa` tests as the examples. In this PR I won't remove the example code, not to clutter the PR more.

I wonder if it shouldn't be the other way around - the examples should be in the example directory and their main purpose should be to be useful as documentation (good readme, etc.). Maybe some additional step in CI that copies examples to the test directories.

If we decide to keep the examples in the examples directory, this PR will make no sense.

szalpal:
> Maybe some additional step in CI that copies examples to the test directories is needed.

This would remove the feature (which is pain in the arse in DALI), that the tests can be run locally - you just need to mount the test dir. I also believe, since not all the examples would constitute a test and not all tests would constitute an example, that such automation isn't really fortunate (we would need to hand-pick them anyway).

> their main purpose should be to be useful as documentation (good readme, etc.)

I believe, that this still can be achieved. Nothing stops us from writing a test like we would write an example. With the advantage, that we can verify correctness of such test-example.

From my perspective, it's going to be easier both for the user and for our infra to have it this way

qa/L0_resnet50_trt/model_repository/ensemble_dali_resnet50/config.pbtxt
```diff
@@ -0,0 +1,66 @@
...
+  {
+    name: "OUTPUT"
+    data_type: TYPE_FP32
+    dims: [ 1000 ]
+  }
+]
+ensemble_scheduling {
+  step [
+    {
+      model_name: "dali"
```
mzient, line 42: This name is rather generic. What's the scope of it? What would happen if there was another ensemble with a model called "dali"?

qa/L0_resnet50_trt/client.py
```diff
@@ -0,0 +1,98 @@
+# The MIT License (MIT)
```
szalpal, line 1: This file is a copy

qa/L0_resnet50_trt/model_repository/dali/config.pbtxt
```diff
@@ -0,0 +1,39 @@
+# The MIT License (MIT)
```
szalpal, line 1: This file is a copy

qa/L0_resnet50_trt/model_repository/ensemble_dali_resnet50/config.pbtxt
```diff
@@ -0,0 +1,66 @@
+# The MIT License (MIT)
```
szalpal, line 1: This file is a copy

qa/L0_resnet50_trt/model_repository/onnx_exporter.py
```diff
@@ -0,0 +1,47 @@
+# The MIT License (MIT)
```
szalpal, line 1: This file is a copy

qa/L0_resnet50_trt/model_repository/resnet50_trt/config.pbtxt
```diff
@@ -0,0 +1,40 @@
+# The MIT License (MIT)
```
szalpal, line 1: This file is a copy

qa/L0_resnet50_trt/model_repository/dali_rn50/config.pbtxt
```diff
@@ -0,0 +1,39 @@
...
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+# FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+name: "dali_rn50"
```
mzient, line 22: :+1:

mzient (approved):

klecki:
> This test is going to be the reference code for the example, due to many reasons: qa/L0_... pattern stipulated by the CI and peculiar setup.sh and test.sh scripts (which probably shouldn't be included in the example) to name two. Thus, when it's merged, I'll think about how to cleanly express qa tests as the examples. In this PR I won't remove the example code, not to clutter the PR more.

Why can't this be done now? I think it should probably be a case of having the additional setup script or whatever that has the paths to the example defined. This way you create a matching directory for example and specify the path to the example that needs to be tested.

Having both the test and example exist in the repo twice (with almost the same code) is wrong for me. And we're just adding more copies, while I don't see any progress to remove duplication.

szalpal:
> Why can't this be done now?

Excuse me if I didn't understand you correctly, but I believe I answered this question in the message you have quoted above:

> In this PR I won't remove the example code, not to clutter the PR more.

szalpal:
Closing for now. We'll pick it up when we figure out how to test the examples in the CI

PR #64, opened 2021-05-20
Fix CPU-only mode, add test for it

szalpal:
One of Triton's modes of operation is that the server works for CPU only (e.g. when somebody runs `docker run` without `--gpus all` option). This PR adds test for this mode of operation and, since it doesn't work, fixes the bug.

Relevant for the review:
* The `L0_identity_cpu` test is copy-pasted `L0_identity` with one difference: DALI pipeline has `device_id=None` set.

### DIFF
--- a/qa/L0_identity_cpu/test.sh
+++ b/qa/L0_identity_cpu/test.sh
@@ -1,0 +1,5 @@
+#!/bin/bash -ex
+
+: ${GRPC_ADDR:=${1:-"localhost:8001"}}
+
+python identity_client.py -u "$GRPC_ADDR"
--- a/src/dali_backend.cc
+++ b/src/dali_backend.cc
@@ -261,7 +261,7 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
       std::vector<IBufferDescr> buffers;
       buffers.reserve(input_buffer_count);
       for (uint32_t buffer_idx = 0; buffer_idx < input_buffer_count; ++buffer_idx) {
-        auto buffer = input.GetBuffer(buffer_idx, device_type_t::CPU, device_id_);
+        auto buffer = input.GetBuffer(buffer_idx, device_type_t::CPU, DetermineDeviceId());
         buffers.push_back(buffer);
       }
       ret.push_back({input.Meta(), std::move(buffers)});
@@ -270,7 +270,8 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
   }
 
   int32_t DetermineDeviceId() {
-    return !CudaStream() ? ::dali::CPU_ONLY_DEVICE_ID : device_id_;}
+    return !CudaStream() ? ::dali::CPU_ONLY_DEVICE_ID : device_id_;
+  }
 
   /**
    * @brief Allocate outputs required by a given request.
@@ -295,7 +296,7 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
       out_meta.type = outputs_info[output_idx].type;
       out_meta.shape = outputs_info[output_idx].shape;
       auto output = response.GetOutput(out_meta);
-      auto buffer = output.AllocateBuffer(outputs_info[output_idx].device, device_id_);
+      auto buffer = output.AllocateBuffer(outputs_info[output_idx].device, DetermineDeviceId());
       outputs[output_idx] = {out_meta, {buffer}};
     }
     return outputs;
--- a/src/dali_backend.cc
+++ b/src/dali_backend.cc
@@ -178,7 +178,7 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
   }
 
   void Execute(const std::vector<TritonRequest>& requests) {
-    DeviceGuard dg(device_id_);
+    DeviceGuard dg(DetermineDeviceId());
     int total_batch_size = 0;
     TimeInterval batch_compute_interval{};
     TimeInterval batch_exec_interval{};
@@ -215,7 +215,7 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
     auto serialized_pipeline = dali_model_->GetModelProvider().GetModel();
     auto max_batch_size = dali_model_->MaxBatchSize();
     auto num_threads = dali_model_->GetModelParamters().GetNumThreads();
-    DaliPipeline pipeline(serialized_pipeline, max_batch_size, num_threads, device_id_);
+    DaliPipeline pipeline(serialized_pipeline, max_batch_size, num_threads, DetermineDeviceId());
     dali_executor_ = std::make_unique<DaliExecutor>(std::move(pipeline));
   }
 
@@ -268,6 +268,9 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
     }
     return ret;
   }
+
+  int32_t DetermineDeviceId() {
+    return !CudaStream() ? ::dali::CPU_ONLY_DEVICE_ID : device_id_;}
 
   /**
    * @brief Allocate outputs required by a given request.
--- a/src/dali_executor/dali_executor.cc
+++ b/src/dali_executor/dali_executor.cc
@@ -50,6 +50,7 @@ void DaliExecutor::SetupInputs(const std::vector<IDescr>& inputs) {
   }
 }
 
+
 IDescr DaliExecutor::ScheduleInputCopy(const IDescr& input) {
   assert(input.buffers.size() > 0);
   IOBufferI* buffer;
@@ -78,6 +79,7 @@ IDescr DaliExecutor::ScheduleInputCopy(const IDescr& input) {
 void DaliExecutor::RunInputCopy() {
   thread_pool_.RunAll();
 }
+
 
 bool DaliExecutor::IsNoCopy(const IDescr& input) {
   return input.buffers.size() == 1 && (input.buffers[0].device == device_type_t::CPU ||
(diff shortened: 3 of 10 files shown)


src/dali_backend.cc
```diff
@@ -220,6 +220,10 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
     return ret;
   }
 
+  int32_t DetermineDeviceId() {
+    return !CudaStream() ? -1 : device_id_;
```
JanuszL, line 224: What is the value of `device_id_` when `!CudaStream()` ?

src/dali_executor/dali_executor.cc
```diff
@@ -65,19 +65,12 @@ IDescr DaliExecutor::ScheduleInputCopy(const IDescr& input) {
   auto descriptor = buffer->get_descr();
   char* dst = reinterpret_cast<char*>(descriptor.data);
   for (auto& buf : input.buffers) {
-    thread_pool_.AddWork(
-        [descriptor, dst, buf](int) {
-          MemCopy(descriptor.device, dst, buf.device, buf.data, buf.size);
-        },
-        buf.size, true);
+    MemCopy(descriptor.device, dst, buf.device, buf.data, buf.size);
```
JanuszL, line 68: Can't you just adjust thread_pool_ contraction and instead of `pipeline_(std::move(pipeline)), thread_pool_(GetNumThreads(), pipeline_.DeviceId(), false) {}` pass -1 as the id when working in the CPU only mode?
DALI does that.

src/dali_executor/dali_pipeline.h
```diff
@@ -160,9 +164,11 @@ class DaliPipeline {
   }
 
   void ReleaseStream() {
+    if (no_gpu_)
```
JanuszL, line 167: If the output is on the CPU does backend still works on stream?

szalpal: IIUC, you're referring to the output of DALI. It might, if the Triton requested GPU output and DALI provides CPU. Since we get to know Triton's requested output in runtime, we need to have the stream ready just in case

src/dali_backend.cc
szalpal: There's no clear explanation for this in Triton docs (the `device_id_` in this case comes from `backend_model_instance.h`). When I logged it, it was `0`, so I suppose we can't trust this. However, from Triton's docs:
```c++
  // Returns the stream associated with this instance that can be used
  // for GPU<->CPU memory transfers. Returns nullptr if GPU support is
  // disabled or if this instance is not executing on a GPU.
  cudaStream_t CudaStream() { return stream_; }
```
Thus `CudaStream` seems it can be trusted in determining, whether the GPU is available

src/dali_executor/dali_executor.cc
szalpal: I tried passing `-1` (ref. `DetermineDeviceId()`), but it still failed with "failed to open `libnvidia-ml.so`". What I see, is that DALI passes `-99999` rather than `-1` in this case. I'll try it out, however it made me thinking, that tightly coupling `dali_backend` code with DALI might length the bugfix cycle, if there's any fix required in DALI code base. I'll discuss it with @banasraf

src/dali_backend.cc
JanuszL: Understood

JanuszL, line 224:
```suggestion
    return !CudaStream() ? ::dali::CPU_ONLY_DEVICE_ID : device_id_;
```

JanuszL (dismissed):

qa/L0_identity_cpu/test.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -ex
+
```
JanuszL, line 2: How about:
```
  export LD_LIBRARY_PATH=""
  export PATH=${PATH/cuda/}
```
just in case?

src/dali_executor/dali_pipeline.h
```diff
@@ -76,7 +76,8 @@ class DaliPipeline {
       serialized_pipeline_(serialized_pipeline),
       max_batch_size_(max_batch_size),
       num_threads_(num_threads),
-      device_id_(device_id) {
+      device_id_(device_id),
+      no_gpu_(device_id_ < 0) {
```
banasraf, line 80: Maybe a `NoGpu()`/`no_gpu()` method instead of a separate field?

szalpal: Done

qa/L0_identity_cpu/test.sh
szalpal: I don't know if I can set `LD_LIBRARY_PATH` to empty. It's mainly because when running tritonserver image, the `LIB_LIBRARY_PATH` is set. I guess emptying it might broke something (though I'm not sure about this)

src/dali_backend.cc
szalpal: I'd prefer to retain the `-1`. I believe this code shouldn't really depend on DALI, cause it's heavily dependent on Triton. I'd prefer having it separated, just in case any changes are needed

qa/L0_identity_cpu/test.sh
JanuszL:
```
export LD_LIBRARY_PATH=${LD_LIBRARY_PATH/cuda/}
``?

File: src/dali_backend.cc (hunk 1 above)
Comment (JanuszL) [reply src/dali_backend.cc:224]: Btu this device ID is for DALI not for the rest TRITON.

File: qa/L0_identity_cpu/test.sh (hunk 4 above)
Comment (szalpal) [reply qa/L0_identity_cpu/test.sh:2]: hmmm, I don't fully understand also the purpose of this?

File: src/dali_backend.cc (hunk 1 above)
Comment (szalpal) [reply src/dali_backend.cc:224]: I'd prefer keeping `dali_backend.cc` separated from DALI's API. We do use `DeviceGuard` here, but that can be easily replaced. And I'd prefer to have all DALI API details encaplusated inside `DaliPipeline` class. I'll discuss it with @banasraf

File: qa/L0_identity_cpu/test.sh (hunk 4 above)
Comment (JanuszL) [reply qa/L0_identity_cpu/test.sh:2]: To make sure that there is not cuda available just in case.

File: src/dali_backend.cc (hunk 1 above)
Comment (JanuszL) [reply src/dali_backend.cc:224]: Well, if the backend used DALI then you cannot separate it. I assume you may need to use DetermineDeviceId valu inside DALI one day so I would prefer to use proper values.

Comment (banasraf) [reply src/dali_backend.cc:224]: I don't see any reason to limit our use of DALI's API. We use multiple parts of it across all of the backend (e.g. TensorListShape is used everywhere). Using DALI stuff just makes coding so much easier (DeviceBuffer, DeviceGuard, CUDAStream, UniqueHandle....) and I don't see any value in treating DALI backend as something separate from DALI.

Comment (szalpal) [reply src/dali_backend.cc:224]: Sounds fair, I'll apply the change

File: qa/L0_identity_cpu/test.sh (hunk 4 above)
Comment (szalpal) [reply qa/L0_identity_cpu/test.sh:2]: OK, so this `test.sh` file is not the place to check it. Without dwelling into details about how the testing is organised, neither `setup.sh` or `test.sh` is called within the docker image containing the server.

The only way to double check CPU only mode at this stage would be to use [client API](https://github.com/triton-inference-server/client/blob/main/src/python/library/tritonclient/grpc/__init__.py) to ask the server. However, it seems, that client API doesn't expose such functionality.

Comment (JanuszL) [reply qa/L0_identity_cpu/test.sh:2]: Right

**Review by JanuszL — APPROVED** [review]

File: src/dali_backend.cc [hunk 6]
```diff
@@ -261,14 +261,18 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+        auto buffer = input.GetBuffer(buffer_idx, device_type_t::CPU, DetermineDeviceId());
         buffers.push_back(buffer);
       }
       ret.push_back({input.Meta(), std::move(buffers)});
     }
     return ret;
   }
 
+  int32_t DetermineDeviceId() {
+    return !CudaStream() ? ::dali::CPU_ONLY_DEVICE_ID : device_id_;
```
mzient, line 273: Why does it depend on the stream? Why not put `CPU_ONLY_DEVICE_ID` or -1 in `device_id_`?

src/dali_backend.cc
```diff
@@ -261,14 +261,18 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
-        auto buffer = input.GetBuffer(buffer_idx, device_type_t::CPU, device_id_);
+        auto buffer = input.GetBuffer(buffer_idx, device_type_t::CPU, DetermineDeviceId());
         buffers.push_back(buffer);
       }
       ret.push_back({input.Meta(), std::move(buffers)});
     }
     return ret;
   }
 
+  int32_t DetermineDeviceId() {
```
mzient, line 272:
```suggestion
  int32_t GetDeviceId() {
```
or (better?)
```suggestion
  int32_t GetDaliDeviceId() {
```
Determine suggests it's done once and then the device id is know - also, the name suggests complex logic.

src/dali_backend.cc
banasraf:
```
// Returns the stream associated with this instance that can be used
// for GPU<->CPU memory transfers. Returns nullptr if GPU support is
// disabled or if this instance is not executing on a GPU.
cudaStream_t CudaStream() { return stream_; }
```

banasraf: Part of triton API

banasraf (approved):

PR #74, opened 2021-05-28
Support chunked output buffers

banasraf:
Ouptut can now have multiple buffers. In that case, DALI pipeline will copy  output to an intermediate buffer. 

### DIFF
--- a/src/dali_executor/dali_executor.cc
+++ b/src/dali_executor/dali_executor.cc
@@ -55,9 +55,9 @@ IDescr DaliExecutor::ScheduleInputCopy(const IDescr& input) {
   assert(input.buffers.size() > 0);
   IOBufferI* buffer;
   if (input.buffers[0].device == device_type_t::CPU) {
-    buffer = &cpu_buffers_[input.meta.name];
+    buffer = &cpu_buffers_[input.meta.name + "_inp"];
   } else {
-    buffer = &gpu_buffers_[input.meta.name];
+    buffer = &gpu_buffers_[input.meta.name + "_inp"];
   }
   size_t size = 0;
   for (auto& buf : input.buffers)
@@ -106,14 +106,38 @@ std::vector<OutputInfo> DaliExecutor::Run(const std::vector<IDescr>& inputs) {
 
 void DaliExecutor::PutOutputs(const std::vector<ODescr>& outputs) {
   for (uint32_t output_idx = 0; output_idx < outputs.size(); ++output_idx) {
-    ENFORCE(outputs[output_idx].buffers.size() == 1,
-            "Ouptut can be copied only to a single buffer");
-    auto buffer = outputs[output_idx].buffers[0];
-    auto data = buffer.data;
-    auto device_type = buffer.device;
-    pipeline_.PutOutput(data, output_idx, device_type);
+    if (outputs[output_idx].buffers.size() == 1) {
+      auto buffer = outputs[output_idx].buffers[0];
+      pipeline_.PutOutput(buffer.data, output_idx, buffer.device);
+    } else {
+      const auto& name = outputs[output_idx].meta.name;
+      const auto& out_buffers = outputs[output_idx].buffers;
+      size_t size = 0;
+      for (auto& out_buff : out_buffers) {
+        size += out_buff.size;
+      }
+      IOBufferI* interm_buffer;
+      if (pipeline_.GetOutputDevice(output_idx) == device_type_t::CPU) {
+        interm_buffer = &cpu_buffers_[name + "_out"];
+      } else {
+        interm_buffer = &gpu_buffers_[name + "_out"];
+      }
+      interm_buffer->resize(size);
+      auto interm_descr = interm_buffer->get_descr();
+      pipeline_.PutOutput(interm_descr.data, output_idx, interm_descr.device);
+      char* src = reinterpret_cast<char*>(interm_descr.data);
+      for (auto& buf : out_buffers) {
+        thread_pool_.AddWork(
+            [src, buf, interm_descr](int) {
+              MemCopy(buf.device, buf.data, interm_descr.device, src, buf.size);
+            },
+            buf.size, false);  // deferred execution
+        src += buf.size;
+      }
+    }
   }
   pipeline_.SyncOutputStream();
+  thread_pool_.RunAll();
 }
 
 }}}  // namespace triton::backend::dali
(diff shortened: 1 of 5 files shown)


src/dali_executor/dali_executor.cc
```diff
@@ -106,14 +106,38 @@ std::vector<OutputInfo> DaliExecutor::Run(const std::vector<IDescr>& inputs) {
...
+            [src, buf, interm_descr](int) {
+              MemCopy(buf.device, buf.data, interm_descr.device, src, buf.size);
+            },
+            buf.size, false);  // deferred execution
+        src += buf.size;
+      }
+    }
   }
   pipeline_.SyncOutputStream();
+  thread_pool_.RunAll();
```
mzient, line 140: Won't hang on GPU? You haven't scheduled the copies (they are deferred!) but wait for them to complete?

src/dali_executor/dali_executor.cc
```diff
@@ -106,14 +106,38 @@ std::vector<OutputInfo> DaliExecutor::Run(const std::vector<IDescr>& inputs) {
...
+        interm_buffer = &gpu_buffers_[name + "_out"];
+      }
+      interm_buffer->resize(size);
+      auto interm_descr = interm_buffer->get_descr();
+      pipeline_.PutOutput(interm_descr.data, output_idx, interm_descr.device);
+      char* src = reinterpret_cast<char*>(interm_descr.data);
+      for (auto& buf : out_buffers) {
+        thread_pool_.AddWork(
+            [src, buf, interm_descr](int) {
+              MemCopy(buf.device, buf.data, interm_descr.device, src, buf.size);
```
mzient, line 132: stream?

banasraf: Added

src/dali_executor/dali_executor.cc
banasraf: Should be fixed.  Now there is a WaitForCopies function that first waits for the thread_pool and then for the stream.

src/dali_executor/dali_executor.h
```diff
@@ -61,16 +61,23 @@ class DaliExecutor {
   void SetupInputs(const std::vector<IDescr>& inputs);
 
   /**
-   * @brief Schedule a copy off all buffers within input IDescr to a continuous buffer.
-   *        The copy will be performed after calling RunInputCopy().
+   * @brief Schedule a copy of all buffers within input IDescr to a continuous buffer.
+   *        Call RunInputCopy() to wait for the copy to finish.
```
mzient, line 65:
```suggestion
   *        Call WaitForCopies() to wait for the copy to finish.
```
?

src/dali_executor/dali_executor.h
```diff
@@ -61,16 +61,23 @@ class DaliExecutor {
...
    * @return IDecr to the new, continuous, buffer.
    */
   IDescr ScheduleInputCopy(const IDescr& buffers);
 
   /**
-   * @brief Run copies scheduled by ScheduleInputCopy and wait for them to finish.
+   * @brief Schedule a copy to a chunked output through an intermediate buffer.
+   *        Call RunInputCopy() to wait for the copy to finish.
```
mzient, line 72:
```suggestion
   *        Call WaitForCopies() to wait for the copy to finish.
```

src/dali_executor/executor.test.cc
```diff
@@ -29,15 +29,49 @@
 
 namespace triton { namespace backend { namespace dali { namespace test {
 
+template<typename T, typename Op>
+void pointwise_compare(const std::vector<OBufferDescr> &obuffers,
+                       const std::vector<std::vector<T>> &ibuffers, size_t inp_size, const Op &op) {
```
mzient, line 34: This name is not very informative.
```suggestion
void coalesced_compare(const std::vector<OBufferDescr> &obuffers,
                       const std::vector<std::vector<T>> &ibuffers, size_t inp_size, const Op &op) {
```

src/dali_executor/dali_executor.h
banasraf: fixed

src/dali_executor/dali_executor.h
banasraf: fixed

src/dali_executor/executor.test.cc
```diff
@@ -29,15 +29,49 @@
...
+  size_t inp_i = 0;
+  size_t out_buff_i = 0;
+  size_t out_i = 0;
+  std::vector<T> obuffer;
+  for (size_t i = 0; i < inp_size; ++i) {
+    if (inp_i == ibuffers[inp_buff_i].size()) {
+      inp_i = 0;
+      inp_buff_i++;
+    }
+    if (out_i == obuffers[out_buff_i].size / sizeof(T)) {
```
mzient, line 45: Is it by any chance possible that a single value would be split?

banasraf: No. The use case for chunked outputs is the the dynamic batching. Each of the outputs buffer will be a response for a single request, so, in fact, not even a sample can be split.

mzient (approved):

src/dali_executor/dali_executor.cc
```diff
@@ -55,29 +55,59 @@ IDescr DaliExecutor::ScheduleInputCopy(const IDescr& input) {
   assert(input.buffers.size() > 0);
   IOBufferI* buffer;
   if (input.buffers[0].device == device_type_t::CPU) {
-    buffer = &cpu_buffers_[input.meta.name];
+    buffer = &cpu_buffers_[input.meta.name + "_inp"];
```
klecki, line 58: It might be better to have a GetCpuInputBuffer or something rather than use a magic suffix string.

klecki: It can even be `IOBufferI* GetInputBuffer(IDescr &, device_type_t);`

src/dali_executor/dali_executor.cc
```diff
@@ -55,29 +55,59 @@ IDescr DaliExecutor::ScheduleInputCopy(const IDescr& input) {
...
+void DaliExecutor::ScheduleOutputCopy(const ODescr& output, int output_idx) {
+  const auto& name = output.meta.name;
+  const auto& out_buffers = output.buffers;
+  size_t size = 0;
+  for (auto& out_buff : out_buffers) {
+    size += out_buff.size;
+  }
+  IOBufferI* interm_buffer;
+  if (pipeline_.GetOutputDevice(output_idx) == device_type_t::CPU) {
+    interm_buffer = &cpu_buffers_[name + "_out"];
```
klecki, line 89: Same here.

banasraf: Added GetInputBuffer and GetOutputBuffer

src/dali_executor/dali_executor.cc
banasraf: Added GetInputBuffer and GetOutputBuffer

klecki (approved):

PR #76, opened 2021-06-01
Enable dynamic batching.

banasraf:
Enabling dynamic batching, which means merging requests that come sufficiently close in time, to increase utilization of DALI pipeline.  

It required refactoring some of the request handling functions. The executor was already prepared to work in such case. 

First measurements show that the change can improve perf significantly in some cases.
I tested the L0_DALI_GPU_ensemble setup with quite small requests (4 samples, each ~10kB), changing the request concurrency from 4 to 10. 
With the current version there was no significant change in throughput when increasing the concurrency (as expected).
With the dynamic batching support the throughput increased by **40% (5900 -> 8370 inf/sec)**, when increasing request concurrency from 4 to 10.

For low concurrency levels, the change does not affect the performance.


qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -51,57 +61,67 @@ def _get_input(batch, name):
...
+  def test_infer(self, batches, it):
     assert(len(batches) == len(self.input_names))
     if (len(batches) > 1):
       for b in batches:
         assert b.shape[0] == batches[0].shape[0]
-    inputs = [self._get_input(batch, name) for batch, name in zip(batches, self.input_names)]
     outputs = [t_client.InferRequestedOutput(name) for name in self.output_names]
-    results = self.client.infer(model_name=self.model_name, inputs=inputs, outputs=outputs)
-    return [results.as_numpy(name) for name in self.output_names]
+    inputs = [self._get_input(batch, name) for batch, name in zip(batches, self.input_names)]
```
JanuszL, line 70: Why switching lines here?

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -51,57 +61,67 @@ def _get_input(batch, name):
...
-        assert out.shape == ref_out.shape
-        if not np.allclose(out, ref_out, atol=eps):
-          print("Test failure in iteration", it)
-          print("Output", out_i)
-          print("Expected:\n", ref_out)
-          print("Actual:\n", out)
-          assert False
-      print('PASS iteration:', it)
+    for pack in grouper(self.concurrency, enumerate(generator)):
+      with ThreadPoolExecutor(max_workers=self.concurrency+4) as executor:
```
JanuszL, line 78: Why `+4`?

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -51,57 +61,67 @@ def _get_input(batch, name):
...
+        results_f = [executor.submit(self.test_infer, data, it) for it, data in pack]
+        for future in as_completed(results_f):
```
JanuszL, line 80: Do this line needs to be in `with ThreadPoolExecutor` scope?

src/dali_executor/utils/utils.h
```diff
@@ -76,6 +78,31 @@ TensorListShape<Dims> cat_list_shapes(const Container &shapes) {
...
 }
 
+template<typename Container, int Dims = -1>
+std::vector<TensorListShape<Dims>> split_list_shape(const TensorListShape<Dims> &shape,
+                                                    const Container &batch_sizes) {
+  size_t nresults = 0;
+  int64_t nsamples = 0;
+  for (const auto &bs : batch_sizes) {
+    nsamples += bs;
+    nresults++;
```
JanuszL, line 88: Do you expect the `Container` to support range iteration but no `size` method?

src/dali_backend.cc
```diff
@@ -180,33 +185,26 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+    for (auto& request : requests) {
+      responses.push_back(TritonResponse::New(request));
+    }
+    ProcessingMeta proc_meta{};
+    TritonError error{};
+    try {
+      proc_meta = ProcessRequests(requests, responses);
+    } catch (...) { error = ErrorHandler(); }
+    for (auto& response : responses) {
+      SendResponse(std::move(response), TritonError::Copy(error));
```
JanuszL, line 201:
```suggestion
      for (auto& response : responses) {
        SendResponse(std::move(response), TritonError::Copy(error));
```

src/dali_backend.cc
```diff
@@ -234,39 +232,61 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
-  /** Run inference for a given \p request and prepare a response. */
-  RequestMeta ProcessRequest(TritonResponseView response, TritonRequestView request) {
-    RequestMeta ret;
-
-    auto dali_inputs = GenerateInputs(request);
-    ret.batch_size = dali_inputs[0].meta.shape.num_samples();  // Batch size is expected to be the
-                                                               // same in every input
+  /**
+   * @brief Run inference for a given \p request and prepare a response.
+   * @return computation time interval
```
JanuszL, line 237:
```suggestion
   * @return computation time interval and total batch size
```

src/dali_backend.cc
```diff
@@ -234,39 +232,61 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+   */
+  ProcessingMeta ProcessRequests(const std::vector<TritonRequest>& requests,
+                                 const std::vector<TritonResponse>& responses) {
+    ProcessingMeta ret{};
+    auto inputs_info = GenerateInputs(requests);
```
JanuszL, line 242: I wonder if there should be the limit of requests number we can run at once. Assume we receive 2 samples in one requests, and then 1000 in the next one. I don't think we should adjust to handle this 1000 at once, and have a hard limit not to blow the memory consumption. Like 500 at most and slit requests if needed.

banasraf: Total batch size will not exceed the `max_batch_size` that is defined in the model config.

qa/L0_DALI_GPU_ensemble/client.py
banasraf: Not an expert on that, but looking at the documentation, `as_completed` must be called under with ThreadPoolExecutor

JanuszL: Sure

qa/L0_DALI_GPU_ensemble/client.py
banasraf: leftovers

qa/L0_DALI_GPU_ensemble/client.py
banasraf: I wanted to make sending requests as fast as possible, so they will be caught together by dynamic batcher. Increasing number of threads seemed to improve that, but after I checked again I don't think it helped at all.

src/dali_executor/utils/utils.h
banasraf: changed to use .size()

src/dali_backend.cc
banasraf: done

src/dali_backend.cc
```diff
@@ -234,39 +232,61 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+        } else {
+          ENFORCE(meta.shape.num_samples() == reqs_batch_sizes[ri],
+                  "Each input in a request must have the same batch size.");
+        }
       }
-      ret.push_back({input.Meta(), std::move(buffers)});
     }
-    return ret;
+    for (const auto& descrs : input_map) {
+      IDescr i_descr = cat_io_descriptors(descrs.second);
```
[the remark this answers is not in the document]
banasraf: I removed it and do it directly when processing requests. I realized that `TensorListShape` has `append` method which made it easier

src/dali_backend.cc
```diff
@@ -278,9 +298,18 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+                                      const std::vector<int>& batch_sizes,
                                       const std::vector<OutputInfo>& outputs_info) {
-    uint32_t output_cnt = request.OutputCount();
+    assert(requests.size() > 0);
+    assert(requests.size() == responses.size());
+    assert(requests.size() == batch_sizes.size());
+    uint32_t output_cnt = requests[0].OutputCount();
+    for (auto& req : requests) {
+      ENFORCE(output_cnt == req.OutputCount(),
+              "All of the requests must require the same number of outputs.");
```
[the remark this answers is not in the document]
banasraf: done

src/dali_backend.cc
JanuszL: 👍

src/dali_backend.cc
```diff
@@ -252,12 +252,16 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
     return ret;
   }
 
-  /** @brief Generate descriptors of inputs provided by a given request. */
+  /**
+   * @brief Generate descriptors of inputs provided by given \p requests
+   * @return input descriptors and batch size of each request
+  */
   InputsInfo GenerateInputs(const std::vector<TritonRequest>& requests) {
+    std::cout << requests.size() << std::endl;
```
JanuszL, line 260: ?

JanuszL (dismissed):

src/dali_backend.cc
banasraf: removed

JanuszL (dismissed):

src/dali_backend.cc
```diff
@@ -234,70 +232,119 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+                        output_indices.size(), ")."));
     std::vector<ODescr> outputs(output_cnt);
     outputs.reserve(output_cnt);
-    for (uint32_t i = 0; i < output_cnt; ++i) {
-      auto name = request.OutputName(i);
-      int output_idx = output_indices.at(name);
+    for (const auto& out_index : output_indices) {
+      auto name = out_index.first;
+      int output_idx = out_index.second;
+      auto shapes = split_list_shape(outputs_info[output_idx].shape, batch_sizes);
```
klecki, line 336: I assume that we execute those merged batches with DALI, that means we need to translate/slice it back for triton output?

src/dali_backend.cc
```diff
@@ -180,33 +185,26 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
-      end_timer_ns(req_exec_interval);
-      ReportStats(requests[i], req_exec_interval, request_meta.compute_interval, !error);
-      SendResponse(std::move(response), std::move(error));
-
-      total_batch_size += request_meta.batch_size;
+    TimeInterval exec_interval{};
+    start_timer_ns(exec_interval);
+    std::vector<TritonResponse> responses;
+    responses.reserve(requests.size());
+    for (auto& request : requests) {
```
klecki, line 192: To match the ProcessRequests (with the GenerateInputs/AllocateOutputs) and SendResposnes I would suggest some AllocateRepsonses or something similar.

src/dali_backend.cc
```diff
@@ -234,70 +232,119 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
...
+      ENFORCE(request.InputCount() == input_cnt,
+              "Each request must provide the same number of inputs.");
+      for (uint32_t input_idx = 0; input_idx < input_cnt; ++input_idx) {
+        auto input = request.InputByIdx(input_idx);
+        auto input_byte_size = input.ByteSize();
+        auto input_buffer_count = input.BufferCount();
+        auto meta = input.Meta();
+        auto& idescr = input_map[meta.name];
+        for (uint32_t buffer_idx = 0; buffer_idx < input_buffer_count; ++buffer_idx) {
+          auto buffer = input.GetBuffer(buffer_idx, device_type_t::CPU, GetDaliDeviceId());
```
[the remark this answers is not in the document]
banasraf: It is preferred and since I cannot check what device is the ExternalSource using, I just assume CPU

src/dali_backend.cc
banasraf: Yes. DaliExecutor handles 'chunked outputs', meaning the output can be copied to non-contiguous buffers.

qa/L0_DALI_GPU_ensemble/model_repository/dali_1/config.pbtxt
```diff
@@ -53,3 +53,8 @@ output [
     dims: [ -1 ]
   }
 ]
+
+dynamic_batching {
```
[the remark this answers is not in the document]
banasraf: This enables dynamic batching - the backend will be provided with multiple requests at once **if** they arrive at the same time. 
With `-c` we actually utilize this option and without `-c` we send requests one by one, so the flow is the same as without `dynamic_batching`. 
That means in the tests (I run both, with `-c` and without) I check dynamic batching, but also the sequential flow.

src/dali_executor/utils/utils.h
```diff
@@ -24,6 +24,8 @@
 #define DALI_BACKEND_UTILS_UTILS_H_
 
 #include <cuda_runtime_api.h>
+#include "src/dali_executor/utils/dali.h"
```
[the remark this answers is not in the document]
klecki: I'm blind

src/dali_backend.cc
banasraf: done

src/dali_backend.cc
```diff
@@ -180,33 +185,26 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
   void Execute(const std::vector<TritonRequest>& requests) {
     DeviceGuard dg(GetDaliDeviceId());
     int total_batch_size = 0;
```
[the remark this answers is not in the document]
banasraf: removed

src/dali_executor/utils/utils.h
```diff
@@ -76,6 +78,30 @@ TensorListShape<Dims> cat_list_shapes(const Container &shapes) {
...
+                      " samples to list shapes of total ", nsamples, " samples."));
+  std::vector<TensorListShape<Dims>> result(nresults);
+  int ri = 0;
+  int ti = 0;
+  for (const auto &bs : batch_sizes) {
+    TensorListShape<Dims> res_shape(bs, shape.sample_dim());
+    for (int64_t i = 0; i < bs; ++i) {
+      res_shape.set_tensor_shape(i, shape.tensor_shape_span(ti++));
+    }
+    result[ri++] = res_shape;
```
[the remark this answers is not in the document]
banasraf: used `move`

src/dali_executor/utils/utils.h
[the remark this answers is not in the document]
banasraf: no

src/dali_executor/utils/utils.h
```diff
@@ -76,6 +78,30 @@ TensorListShape<Dims> cat_list_shapes(const Container &shapes) {
   return result;
 }
 
+template<typename Container, int Dims = -1>
+std::vector<TensorListShape<Dims>> split_list_shape(const TensorListShape<Dims> &shape,
+                                                    const Container &batch_sizes) {
+  size_t nresults = batch_sizes.size();
+  int64_t nsamples = 0;
+  for (const auto &bs : batch_sizes) {
```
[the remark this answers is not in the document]
klecki: You left this one with `const auto &`

klecki (approved):

klecki (approved):

PR #82, opened 2021-06-24
Enable using DALI plugins

szalpal:
This PR contains an example of using DALI plugins together with DALI backend. Also, some updates were necessary to make this working:
* added plugin libs loading
* added a parameter, which lists DALI plugins


docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
+# DALI plugin with Triton
```
mzient, line 1:
```suggestion
# DALI plugins with Triton
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+This example shows how to use custom operations implemented for DALI
+(i.e. DALI plugins) with Triton. As an example, we use here a custom DALI
+operation that copies the input to the output. You can find this copy-plugin in
```
mzient, line 5:
```suggestion
(i.e. DALI plugins) with Triton. In this example, we use a dummy DALI
operation that copies the input to the output. You can find this copy plugin in
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+[DALI documentation](https://docs.nvidia.com/deeplearning/dali/master-user-guide/docs/examples/custom_operations/custom_operator/create_a_custom_operator.html).
+
+### Step 1: Obtain the plugin lib
+
+Firstly, you need to code and build DALI plugin. This topic is
```
mzient, line 10:
```suggestion
Firstly, you need to code and build the plugin. This topic is
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+out of scope of this tutorial. You can find all relevant info in
```
mzient, line 11:
```suggestion
out of scope of this tutorial. You can find all relevant information in
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+[DALI documentation](https://docs.nvidia.com/deeplearning/dali/master-user-guide/docs/examples/custom_operations/custom_operator/create_a_custom_operator.html). All you need to do is to follow linked tutorial to obtain `.so` library with your custom plugin.
```
mzient, line 12:
```suggestion
[DALI documentation](https://docs.nvidia.com/deeplearning/dali/master-user-guide/docs/examples/custom_operations/custom_operator/create_a_custom_operator.html). All you need to do is to follow the linked tutorial to obtain `.so` library with your custom plugin.
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Step 2: Serialize DALI pipeline which uses your custom operation
```
mzient, line 14:
```suggestion
### Step 2: Serialize the DALI pipeline which uses your custom operation
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+    plugin_manager.load_library('/path/to/libcustomcopy.so')
+    
+    @dali.pipeline_def(batch_size=3, num_threads=1, device_id=0)
+    def pipe():
+        data = dali.fn.external_source(device="cpu", name="DALI_INPUT_0")
+        cp = dali.fn.custom_copy(data)
+        return cp
+
+### Step 3: Inject plugin lib into tritonserver docker image
```
mzient, line 29:
```suggestion
### Step 3: Inject the plugin lib into tritonserver docker image
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+The `.so` file you have built in the previous step contains the
```
mzient, line 31:
```suggestion
The `.so` file you built in the previous step contains the
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+definition of your operation. Therefore, you need to inject it
+into your tritonserver image, so that DALI would be able to conduct
+this operation. There are multiple ways to achieve that, including
+mounting (`-v` option), copying (`COPY` directive or `docker cp` command),
+pick whichever suits you best.
```
mzient, line 36:
```suggestion
```suggestion
definition of your operation. The file must be available inside your
tritonserver docker image so that DALI can load it and use the newly
defined operation. This can be achieved in several ways, including
mounting (`-v` option) or copying (`COPY` directive or `docker cp` command).
```

> pick whichever suits you best.
sounds a bit too casual

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Here we will assume, that your `.so` library is located at
+`/models/libcustomcopy.so` path in your tritonserver docker image.
```
mzient, line 39:
```suggestion
Here we will assume, that your `.so` library is located at
`/models/libcustomcopy.so` in your tritonserver docker image.
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+### Step 4: Update your config file
+
+DALI Backend will be able to pick up your plugin library, provided
+that you'll point it out in model configuration. In your model repository
+please add a parameter to the DALI model, listing colon-separated plugin libraries:
+
+    parameters: [
+        {
+            key: "plugin_libs"
+            value: { string_value: "/path/to/libplugin1.so:/another/path/libplugin2.so" }
```
mzient, line 50: I find it a bit confusing - previous step says that it assumes some location of the file and now you show _two_ plugins in different location.

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+pick whichever suits you best.
+
+Here we will assume, that your `.so` library is located at
+`/models/libcustomcopy.so` path in your tritonserver docker image.
+
+### Step 4: Update your config file
+
+DALI Backend will be able to pick up your plugin library, provided
+that you'll point it out in model configuration. In your model repository
+please add a parameter to the DALI model, listing colon-separated plugin libraries:
```
mzient, line 45: This is awkward.
```suggestion
For DALI Backend to be able to use your plugin libraries, you must point to
them in the model configuration file. Add a ``plugin_libs`` parameter to the DALI model,
with the value being a colon-separated list of paths to plugin libraries.
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+    parameters: [
+        {
+            key: "plugin_libs"
+            value: { string_value: "/path/to/libplugin1.so:/another/path/libplugin2.so" }
+        }
+    ]
+
+You can find an example in `model_repository/mydali/config.pbtxt` file.
+
+### Step 5: Enjoy custom DALI plugin in Triton
```
mzient, line 56:
```suggestion
### Step 5: Enjoy your custom DALI plugin in Triton
```

docs/examples/dali_plugin/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+That's all. You can now run your server
+
+## Remember
+
+As always in DALI Backend case, remember that `dali.fn.external_source`'s `name` parameter must match
+with input name provided in `config.pbtxt` file.
```
mzient, line 63:
```suggestion
the input name provided in `config.pbtxt` file. 
```

docs/examples/dali_plugin/custom_copy_client.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    lengths = list(map(lambda x, arr=arrays: arr[x].shape[0], [x for x in range(len(arrays))]))
+    max_len = max(lengths)
+    arrays = list(map(lambda arr, ml=max_len: np.pad(arr, ((0, ml - arr.shape[0]))), arrays))
+    for arr in arrays:
+        assert arr.shape == arrays[0].shape, "Arrays must have the same shape"
+    return np.stack(arrays)
+
+
+def batcher(dataset, max_batch_size, n_iterations=-1):
```
mzient, line 60: That name is positively foreign. Batcher is a machine for preparing a batch of concrete :)
There are mere 1870 google results for `batcher pytorch`.

docs/examples/dali_plugin/custom_copy_pipeline.py
```diff
@@ -0,0 +1,47 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2020-2021 NVIDIA CORPORATION
```
mzient, line 3:
```suggestion
# Copyright (c) 2020-2021 NVIDIA CORPORATION & AFFILIATES
```

docs/examples/dali_plugin/custom_copy_client.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
+
+# The MIT License (MIT)
+#
+# Copyright (c) 2020 NVIDIA CORPORATION
```
mzient, line 5:
```suggestion
# Copyright (c) 2020-2021 NVIDIA CORPORATION & AFFILIATES
```

src/dali_executor/utils/utils.h
```diff
@@ -20,8 +20,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
```
mzient, line 22: update license header

src/utils/utils.h
```diff
@@ -0,0 +1,70 @@
...
+ * @return
+ */
+inline std::vector<std::string> split(const std::string& str, const std::string& delimiter = ":") {
+  std::vector<std::string> ret;
+  size_t start = 0, end = 0;
+  do {
+    end = str.find(delimiter, start);
+    auto substring = str.substr(start, end - start);
+    if (!substring.empty()) {
+      ret.emplace_back(substring);
```
mzient, line 45:
```suggestion
      ret.push_back(std::move(substring));
```

src/utils/utils.h
```diff
@@ -0,0 +1,70 @@
...
+    start = end + delimiter.length();
+  } while (end != std::string::npos);
+  return ret;
+}
+
+
+template<typename T>
+T from_string(const std::string& str) {
+  assert(false);  // Not implemented
+}
```
mzient, line 56: If it's not implemented then don't implement it. Why wait with the error until run time?
```suggestion
template<typename T>
T from_string(const std::string& str);
```

src/utils/utils.test.cc
```diff
@@ -0,0 +1,70 @@
...
+      {"This is ", "a string"},
+  };
+
+  REQUIRE(inputs.size() == ref.size());
+  for (size_t i = 0; i < inputs.size(); i++) {
+    REQUIRE(ref[i] == split(inputs[i]));
+  }
+}
+
+}}}}  // namespace triton::backend::dali::test
```
mzient, line 70: Missing new line at the end of file.

docs/examples/dali_plugin/custom_copy_client.py
szalpal: I used the noun, to reflect that this is not an ordinary function, but a generator. Anyhow, this is a fair point.
@banasraf , what's your opinion? Shall we change this name to `generate_batches`? Or `batch_generator`?

banasraf: No strong opinion. Generally I would prefer verbs for function names, but a generator is kind of different case, so I'm ok with both options

docs/examples/dali_plugin/setup.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
+
+# The MIT License (MIT)
+#
+# Copyright (c) 2021 NVIDIA CORPORATION
```
banasraf, line 5: and affiliates

src/dali_model_instance.h
```diff
@@ -62,7 +62,8 @@ class DaliModelInstance : public ::triton::backend::BackendModelInstance {
     auto serialized_pipeline = dali_model_->GetModelProvider().GetModel();
     auto max_batch_size = dali_model_->MaxBatchSize();
     auto num_threads = dali_model_->GetModelParamters().GetNumThreads();
-    DaliPipeline pipeline(serialized_pipeline, max_batch_size, num_threads, GetDaliDeviceId());
+    DaliPipeline pipeline(serialized_pipeline, max_batch_size, num_threads, GetDaliDeviceId(),
```
banasraf, line 65: What would happen if we load the same plugin .so twice?

docs/examples/dali_plugin/model_repository/mydali/config.pbtxt
```diff
@@ -0,0 +1,46 @@
...
+    dims: [ -1 ]
+  }
+]
+
+parameters: [
+  {
+    key: "plugin_libs"
+    value: { string_value: "/models/libcustomcopy.so" }
+  }
+]
```
banasraf, line 46: newline

docs/examples/dali_plugin/custom_copy_client.py
mzient: I'm fine with both generate_ and _generator. `batcher`- well, as I said, sounds like an industrial machine - I've only found one use of "batcher" in NN context and it wasn't in code written by a native English speaker.

src/dali_model_instance.h
mzient: Nothing. It would likely break, though, if there were two copies of the library loaded from different paths - both would be initialized, but the operator name would clash resulting in weird behavior. If the maps are consistently populated with `insert`, the one loaded first would prevail; if they are populated with `[]` and assignment - the one loaded last would prevail. If there's a mix of the two assignment techniques, there would be a mess, leading to all sorts of weird bugs.

docs/examples/dali_plugin/custom_copy_client.py
szalpal: Went for `batch_generator`

docs/examples/dali_plugin/custom_copy_client.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
```
banasraf, line 1: I fell like this example doesn't need a client script

szalpal: I added a client to actually test, that it works. When we'll have unified client for tests, we could omit this

szalpal:
As we agreed offline, I'll go with the backed-wise setting of the plugins. If another way would be more proper in future, we'll change it.

docs/examples/dali_plugin/custom_copy_client.py
banasraf: So I would add a client when we actually test it (and the client will probably go to a corresponding L0_ directory and not here)

banasraf: I think it would be nice to separate testing files from example files not to confuse readers

szalpal: Done

banasraf (approved):

src/dali_executor/dali_pipeline.h
```diff
@@ -150,6 +150,13 @@ class DaliPipeline {
     return num_threads_;
   }
 
+  static void LoadPluginLibs(const std::vector<std::string>& plugin_paths) {
+    InitDali();
+    for (auto&& path : plugin_paths) {
```
mzient, line 155: Don't create an illusion of being able to move these:
```suggestion
    for (const auto& path : plugin_paths) {
```

src/utils/utils.h
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Splits string according to a delimiter.
+ *
+ * This function doesn't include empty strings in the output.
+ *
+ * @param str String to be split
+ * @param delimiter Delimiter. Won't be included in the output
+ * @return
+ */
+inline std::vector<std::string> split(const std::string& str, const std::string& delimiter = ":") {
```
mzient, line 38: I don't think that ":" is a common enough delimiter as to be used as a default argument to a generic function at C++ level:
```suggestion
inline std::vector<std::string> split(const std::string& str, const std::string& delimiter) {
```

src/parameters.h
```diff
@@ -0,0 +1,116 @@
...
+  template<typename T>
+  T GetParam(const std::string& key, const T& def = T()) const {
+    T result = def;
+    GetMember(key, result);
+    return result;
+  }
+
+  std::vector<std::string> GetPluginNames() const {
+    auto plugin_list = GetParam<std::string>("plugin_libs");
+    return split(plugin_list, ":");
```
mzient, line 92: Can Triton run on Windows? If so, this should be configurable or at the very least OS-dependent. `:` is a valid component of a file path in Windows and indeed an essential part of most (like `C:\`).

src/utils/utils.test.cc
```diff
@@ -0,0 +1,70 @@
...
+      {"This is a string"},
+      {"This is a string"},
+      {"This is a string"},
+      {"Th", "is is a str", "ing"},
+      {"This is ", "a string"},
+  };
+
+  REQUIRE(inputs.size() == ref.size());
+  for (size_t i = 0; i < inputs.size(); i++) {
+    REQUIRE(ref[i] == split(inputs[i]));
```
mzient, line 66:
```suggestion
    REQUIRE(ref[i] == split(inputs[i], ":"));
```

src/utils/utils.h
```diff
@@ -0,0 +1,68 @@
...
+  std::vector<std::string> ret;
+  size_t start = 0, end = 0;
+  do {
+    end = str.find(delimiter, start);
+    auto substring = str.substr(start, end - start);
+    if (!substring.empty()) {
+      ret.emplace_back(std::move(substring));
+    }
+    start = end + delimiter.length();
+  } while (end != std::string::npos);
```
mzient, line 48: This is undefined behavior - you're doing arithmetic between (potentially) `string::npos` and a valid index - it works by coincidence and might as well throw or raise an assertion.

```suggestion
  for (size_t start = 0, len = str.length(); start < len; ) {
    end = str.find(delimiter, start);
    if (end == std::string::npos)
      end = len;  // now end is a correct index
    if (end > start)  // use simpler condition and avoid a call to `substr`
      ret.push_back(str.substr(start, end - start));
    start = end + delimiter.length();
  } while (end != std::string::npos);
```

src/utils/utils.test.cc
mzient: Also, it would be nice to add at least one test with a muti-character delimiter.

src/dali_executor/dali_pipeline.h
```diff
@@ -150,6 +150,13 @@ class DaliPipeline {
...
+      daliLoadLibrary(path.c_str());
```
mzient, line 156: Hmm... I think it's actually a fix for DALI's C API, but it can _throw_....

src/parameters.h
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+ private:
+  template<typename T>
+  void GetMember(const std::string& key, T& value) const {
+    auto key_c = key.c_str();
+    if (params_.Find(key_c)) {
+      std::string string_value{};
```
mzient, line 100:
```suggestion
      std::string string_value;
```

src/parameters.h
szalpal: Triton has a Windows build. How DALI would work there, I honestly do not know. Anyway, added a OS-dependent separator.

src/utils/utils.h
```diff
@@ -0,0 +1,69 @@
...
+ */
+inline std::vector<std::string> split(const std::string& str, const std::string& delimiter) {
+  std::vector<std::string> ret;
+  for (size_t start = 0, len = str.length(); start < len;) {
+    auto end = str.find(delimiter, start);
+    if (end == std::string::npos) {
+      end = len;
+    }
+    if (end > start) {
+      ret.emplace_back(str.substr(start, end - start));
```
mzient, line 46: I've just discovered that there's a substr-like constructor - this will avoid one copy. It's not particularly important in this application (it's no like we load plugins on a time-critical path), but it's a general purpose utility, so...
```suggestion
      ret.emplace_back(str, start, end - start);
```

src/parameters.h
```diff
@@ -0,0 +1,122 @@
...
+#ifndef TRITONDALIBACKEND_PARAMETERS_H
+#define TRITONDALIBACKEND_PARAMETERS_H
+
+#include "src/utils/triton.h"
+#include "src/utils/utils.h"
+
+
+namespace triton { namespace backend { namespace dali {
+
+#if defined(_MSC_VER)
```
mzient, line 32: It's the target OS, not the compiler, we're insterested in. I don't think gcc for windows would define _MSC_VER (though clang might).
```suggestion
#if defined(_WIN32)
```

mzient (approved):

banasraf (approved):

PR #83, opened 2021-06-25
Move TestClient to a common directory

banasraf:
Refactoring to move `TestClient` from L0_DALI_GPU_ensemble test to a common directory so it can be reused in other tests. 


szalpal (approved):

qa/L0_DALI_GPU_ensemble/client.py
```diff
@@ -19,91 +19,22 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-import tritonclient.grpc as t_client
 import numpy as np
-from typing import Sequence
-from itertools import cycle, islice
-from numpy.random import randint
+from dali_backend.test_utils.client import TestClient
```
banasraf, line 23: It relies on the fact that _dali_backend/python_ dir is added to the `PYTHONPATH`

JanuszL (approved):

qa/L0_DALI_GPU_ensemble/client.py
JanuszL: Is this utility going to be shipped with the regular builds?

banasraf: Nope. The only official build we have is the official Triton docker so we don't ship anything more than the backend binary plugged in the server.

PR #91, opened 2021-07-28
Fix time reporting bug

banasraf:
I used a wrong epoch for timestamps so it resulted in a wrong value being reported as `queue` time. 


szalpal (approved):

src/utils/triton.h
```diff
@@ -99,6 +99,19 @@ inline TRITONSERVER_MemoryType to_triton(device_type_t dev) {
   }
 }
 
+struct TimeInterval {
+  int64_t start = 0;
+  int64_t end = 0;
+};
+
+inline void start_timer_ns(TimeInterval &interval) {
+  SET_TIMESTAMP(interval.start);
```
banasraf, line 108: It's a utility provided by the Triton backend lib, so it's better to use it instead of using clock functions by hand.

klecki (approved):

PR #93, opened 2021-07-30
Check external source backend to optimize intermediate buffer allocation.

banasraf:
I check the backend of an external source to allocate an intermediate buffer on the same device. Due to that I can force no_copy mode when setting the input.


awolant (approved):
Maybe removing this as a flag would be useful?

klecki (commented):
As discussed: please add test coverage for cases CPU/GPU Triton input (that needs or needs not to be copied to temporary buffer) and cpu/gpu `device` in external_source.

I think that IsNoCopy needs to be extended to take the device of ES into account as DALI will raise an error if we request no-copy for cross-device inputs.

klecki (approved):
I wonder which one is better, to do, copy to a temporary buffer in backend and use no-copy or let DALI do the copy in case of the one contiguous chunk from trition?

awolant (approved):

klecki (approved):

awolant (approved):

PR #113, opened 2022-01-11
Working around build error from libcudacxx

szalpal:
In latest release, libcudacxx prompts a build error, that a typedef is unused:
```
/cuda/memory_resource:818:13: error: typedef ‘using __view1_t = class cuda::__4::basic_resource_view<_ResourcePointer, _Properties>’ locally defined but not used [-Werror=unused-local-typedefs]
  818 |       using __view1_t = basic_resource_view;
```

This PR turns off this warning. It may be reverted, when the problem in libcudacxx is fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ endif ()  # TRITON_ENABLE_GPU
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-variable -Wno-unused-function")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-variable -Wno-unused-function -Wno-unused-local-typedefs")
 
 if (WERROR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ endif ()  # TRITON_ENABLE_GPU
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-variable -Wno-unused-function")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-variable -Wno-unused-function -Wno-unused-local-typedefs")
```
mzient, line 99: OK, since it unblocks CI - but in the long run, shouldn't we rather treat DALI as a system library? This would suppress all warnings in DALI headers.

mzient (approved):
OK, but see the comment.

JanuszL (approved):

CMakeLists.txt
JanuszL: I'm not sure how feasible it is in the current setup.

PR #115, opened 2022-01-20
Turn `-Werror` off by default

szalpal:
We shall turn `-Werror` off by default.


JanuszL (approved):

awolant (approved):

rmccorm4:
@szalpal was this PR related to Clang13+ new warning flag causing errors when not handled: https://github.com/triton-inference-server/server/issues/4332

szalpal:
@rmccorm4 ,

Not really. I made this change because of two reasons:
1. I believe it is a general good practice to turn `-Werror` off by default. Looking from the perspective of The Project, that includes and builds My Project. Whenever any warning in My Project happens, for example because The Project uses different (older or newer) compiler version, of different compiler, or different version of the dependency of My Project, `-Werror` makes such warning break the compilation.
2. Following the first point here. There has been a brief period when `libcudacxx` introduced a warning in DALI, which propagated to the DALI Backend (the two latter share some parts of code). This unfortunately broke the Triton build, since Triton didn't knew about that warning. Turning `-Werror` off by default is the right thing to do here.

Still, it would be a good thing to turn `-Werror` on for the CI builds, which we actually do.

PR #119, opened 2022-03-08
Prepare automation to benchmark DALI vs Python

szalpal:
This is first part of benchmarking pipelines: DALI vs Python-backend-based. This commit prepares outline of the setup, not automating anything for now. The benchmark is conducted via `perf_analyzer`. I'm assuming, that the inputs to the both pipelines under test are the same.

The final solution would consist of two general functionalities:
1. Benchmarking the pipelines,
2. Verifying output consistency.

Due to nature of the processing and differences in the implementation, we can't assume, that the outputs will be bit-exact. For now I'm focusing on simplifying the manual verification. In the RN50 case, the images are saved to the file, so they can by easily viewed from numpy.

As I stated above, this is the first PR out of a couple, so there are still hard-coded parts, which I will be changing at the later stage.


benchmarks/dali_vs_python/BM_rn50/model_repository/rn50_dali/config.pbtxt
```diff
@@ -0,0 +1,39 @@
...
+    name: "DALI_INPUT_0"
+    data_type: TYPE_UINT8
+    dims: [ -1 ]
+  }
+]
+
+output [
+  {
+    name: "DALI_OUTPUT_0"
+    data_type: TYPE_UINT8
```
banasraf, line 36: Shouldn't RN50 preprocessing output float values?

szalpal: It should, good catch. I'll change it.

benchmarks/dali_vs_python/BM_rn50/model_repository/rn50_dali/pipeline.py
```diff
@@ -0,0 +1,57 @@
...
+    images = dali.fn.resize(images, resize_x=224, resize_y=224)
+    return dali.fn.crop_mirror_normalize(images,
+                                         dtype=types.FLOAT,
+                                         output_layout="HWC",
+                                         crop=(224, 224),
+                                         mean=[0.485 * 255, 0.456 * 255, 0.406 * 255],
+                                         std=[0.229 * 255, 0.224 * 255, 0.225 * 255])
+
+
+@dali.pipeline_def(batch_size=1, num_threads=1, device_id=0)
```
banasraf, line 45: Just a note: we might want to start setting the `batch_size` and `num_threads` values to the desired values here, because in the future we won't specify them in the model config

benchmarks/dali_vs_python/BM_rn50/model_repository/rn50_python/config.pbtxt
```diff
@@ -0,0 +1,39 @@
...
+    name: "PYTHON_INPUT_0"
+    data_type: TYPE_UINT8
+    dims: [ -1 ]
+  }
+]
+
+output [
+  {
+    name: "PYTHON_OUTPUT_0"
+    data_type: TYPE_UINT8
```
banasraf, line 36: float?

benchmarks/dali_vs_python/BM_rn50/run_benchmark.py
```diff
@@ -0,0 +1,126 @@
...
+        'compute_infer': search(r'(?<=compute infer )\d+'),
+        'compute_output': search(r'(?<=compute output )\d+'),
+        'server_latency': search(r'(?<=Avg request latency: )\d+'),
+        'client_latency': search(r'(?<=Avg latency: )\d+'),
+        'client_throughput': search(r'(?<=Throughput: )\d+')
+    }
+    return ret
+
+
+def generate_csv(result_dicts):
```
banasraf, line 73: `perf_analyzer -f file.csv` outputs the resuts to the csv

awolant (dismissed):

benchmarks/dali_vs_python/BM_rn50/model_repository/rn50_dali/config.pbtxt
szalpal: Done

benchmarks/dali_vs_python/BM_rn50/model_repository/rn50_python/config.pbtxt
szalpal: Done

benchmarks/dali_vs_python/BM_rn50/run_benchmark.py
szalpal: That's true, however single `perf_analyzer` run is defined per single batch size. Therefore, I'd still need to call `perf_analyzer` multiple (say 10) times, create 10 CSVs (one per every batch size), open them, parse and write all the data to a new CSV. Judging from this, parsing the console output seems easier. Unless you'd have a better idea? I'm open to any suggestions

benchmarks/dali_vs_python/BM_rn50/model_repository/rn50_dali/pipeline.py
szalpal: But model config only sets the max batch size, right? Also, `batch_size` option in the decorator is in fact a `max_batch_size`. IIUC, the batch size should be actually inferred from the input, so I believe there's no need of setting the desired value here.

About the `num_threads`, in this case whatever is passed to the model config will overwrite the decorator setting. When the configuration is autogenerated, we will still need to default to some setting, and `1` seems to be a good default? Unless you have some other opinion on this?

benchmarks/dali_vs_python/BM_rn50/run_benchmark.py
banasraf: Ok, seems reasonable

banasraf (approved):

awolant (approved):

PR #120, opened 2022-03-15
Fix incorrect test definition for perf_analyer example

szalpal:
This PR solves the problem of the `perf_analyzer_example` test, which has been sometimes failing. The testing conditions weren't correct - they assumed, that the throughput of the pipeline will always be in range `100..999 [ims/s]`. We should allow it to be `0..9999 [ims/s]`, as these are sane values in proper circumstances.

### DIFF
--- a/qa/L0_perf_analyzer_example/test.sh
+++ b/qa/L0_perf_analyzer_example/test.sh
@@ -23,6 +23,6 @@
 
 : ${GRPC_ADDR:=${1:-"localhost:8001"}}
 
-perf_analyzer -m dali -i grpc -u $GRPC_ADDR -b 64 --input-data test_image --shape DALI_INPUT_0:`stat --printf="%s" test_image/DALI_INPUT_0` -f results.txt
+perf_analyzer -m dali -i grpc -u $GRPC_ADDR -b 64 --input-data test_image --shape DALI_INPUT_0:"$(stat --printf='%s' test_image/DALI_INPUT_0)" -f results.txt
 # Test if the perf_analyzer output is in a proper format and the measured times are in a reasonable range.
-[[ `tail -n1 results.txt` =~ ^1,[0-9]{3}\.[0-9],([0-9]{1,6},){10}[0-9]{1,6}$ ]] && echo "Output Correct"
+[[ "$(tail -n1 results.txt)" =~ ^1,[0-9]{1,4}\.[0-9],([0-9]{1,7},){10}[0-9]{1,7}$ ]] && echo "Output Correct"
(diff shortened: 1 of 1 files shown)


qa/L0_perf_analyzer_example/test.sh
```diff
@@ -23,6 +23,6 @@
 
 : ${GRPC_ADDR:=${1:-"localhost:8001"}}
 
-perf_analyzer -m dali -i grpc -u $GRPC_ADDR -b 64 --input-data test_image --shape DALI_INPUT_0:`stat --printf="%s" test_image/DALI_INPUT_0` -f results.txt
+perf_analyzer -m dali -i grpc -u $GRPC_ADDR -b 64 --input-data test_image --shape DALI_INPUT_0:"$(stat --printf='%s' test_image/DALI_INPUT_0)" -f results.txt
 # Test if the perf_analyzer output is in a proper format and the measured times are in a reasonable range.
-[[ `tail -n1 results.txt` =~ ^1,[0-9]{3}\.[0-9],([0-9]{1,6},){10}[0-9]{1,6}$ ]] && echo "Output Correct"
+[[ "$(tail -n1 results.txt)" =~ ^1,[0-9]{1,4}\.[0-9],([0-9]{1,7},){10}[0-9]{1,7}$ ]] && echo "Output Correct"
```
mzient, line 28: Why change backquotes to `$()`?

JanuszL: I prefer `$(`. But this is just my opinion.

JanuszL, line 28:
```suggestion
[[ "$(tail -n1 results.txt)" =~ ^1,[0-9]{1,}\.[0-9],([0-9]{1,},){10}[0-9]{1,}$ ]] && echo "Output Correct"
```
Maybe you should not put the upper bound of the number of occurrences?

szalpal: It is quite discouraged to use backtick notation (e.g. by the [shellcheck](https://github.com/koalaman/shellcheck/wiki/SC2006)). I believe the main reason is that nesting backticks (which need to be escaped in this case) turn into a mess. Anyway, the fact, that backticks are legacy, `$()` is standardized, equivalent and encouraged convinces me to use it :)

szalpal: Makes sense, done!

mzient: Whatever. For simple usages I still think it's cleaner, but I definitely agree with the nesting argument.

mzient (approved):

JanuszL (approved):

PR #122, opened 2022-03-20
Correct the output time unit of ResNet50 example script

laochonlam:
Fixed the output time unit of ResNet50 example script.

### DIFF
--- a/docs/examples/resnet50_trt/client.py
+++ b/docs/examples/resnet50_trt/client.py
@@ -95,4 +95,4 @@ if __name__ == "__main__":
 
     maxs = np.argmax(output0_data, axis=1)
 
-    print("{}ms class: {}".format(latency * 1000, labels_dict[maxs[0]]))
+    print(f"{latency * 1000}ms class: {labels_dict[maxs[0]]}")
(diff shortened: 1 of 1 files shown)


docs/examples/resnet50_trt/client.py
```diff
@@ -95,4 +95,4 @@ def load_image(img_path: str):
 
     maxs = np.argmax(output0_data, axis=1)
 
-    print("{}ms class: {}".format(latency, labels_dict[maxs[0]]))
+    print("{}ms class: {}".format(latency * 1000, labels_dict[maxs[0]]))
```
szalpal, line 98: Hi @laochonlam , thank you for the contribution!

While you are at it, could you change this legacy formatting way to the better one?
```suggestion
    print(f"{latency * 1000}ms class: {labels_dict[maxs[0]]}")
```

JanuszL (approved):

jantonguirao (approved):

docs/examples/resnet50_trt/client.py
laochonlam: Sure!

szalpal (approved):

PR #124, opened 2022-03-30
Add NVTX markers for DALI Backend

szalpal:
This PR adds NVTX markers for profiling DALI Backend with `nsys`. Also, creates a `DALI Backend` domain for the same usage.

Here's roughly, how it looks with the proposed markers:
image

### DIFF
--- a/src/dali_backend.cc
+++ b/src/dali_backend.cc
@@ -35,6 +35,7 @@ extern "C" {
 // should initialize any global state that is intended to be shared
 // across all models and model instances that use the backend.
 TRITONSERVER_Error* TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend) {
+  TimeRange tr("[DALI BE] Initialize", TimeRange::kNavy);
   const char* cname;
   RETURN_IF_ERROR(TRITONBACKEND_BackendName(backend, &cname));
   std::string name(cname);
@@ -90,6 +91,7 @@ TRITONSERVER_Error* TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend) {
 // using TRITONBACKEND_BackendSetState. The backend must free this
 // state and perform any other global cleanup.
 TRITONSERVER_Error* TRITONBACKEND_Finalize(TRITONBACKEND_Backend* backend) {
+  TimeRange tr("[DALI BE] Finalize", TimeRange::kNavy);
   void* vstate;
   RETURN_IF_ERROR(TRITONBACKEND_BackendState(backend, &vstate));
   std::string* state = reinterpret_cast<std::string*>(vstate);
@@ -106,6 +108,7 @@ TRITONSERVER_Error* TRITONBACKEND_Finalize(TRITONBACKEND_Backend* backend) {
 // should initialize any state that is intended to be shared across
 // all instances of the model.
 TRITONSERVER_Error* TRITONBACKEND_ModelInitialize(TRITONBACKEND_Model* model) {
+  TimeRange tr("[DALI BE] ModelInitialize", TimeRange::kNavy);
   const char* cname;
   RETURN_IF_ERROR(TRITONBACKEND_ModelName(model, &cname));
   std::string name(cname);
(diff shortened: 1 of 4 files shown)


src/dali_executor/utils/utils.h
```diff
@@ -129,7 +151,7 @@ struct TimeRange {
     att.messageType = NVTX_MESSAGE_TYPE_ASCII;
     att.message.ascii = name.c_str();
 
-    nvtxRangePushEx(&att);
+    nvtxDomainRangePushEx(detail::NvtxDomain::Get(), &att);
```
JanuszL, line 154: I think the signature should be:
```
TimeRange(const std::string &name, const uint32_t rgb = kPantyPink) { // NOLINT```

File: src/dali_model_instance.cc [hunk 2]
```diff
@@ -109,19 +109,33 @@ std::vector<TritonResponse> DaliModelInstance::CreateResponses(
   return responses;
 }
 
+
 ProcessingMeta DaliModelInstance::ProcessRequests(const std::vector<TritonRequest>& requests,
                                                   const std::vector<TritonResponse>& responses) {
   ProcessingMeta ret{};
+
+  TimeRange tr_gi("[DALI BE] GenerateInputs", 0x488FB1);
```
JanuszL, line 117:
```suggestion
  TimeRange tr_proc"[DALI BE] ProcessRequests", kRed1);
  TimeRange tr_gi("[DALI BE] GenerateInputs", 0x488FB1);
```

src/dali_model_instance.cc
```diff
@@ -109,19 +109,33 @@ std::vector<TritonResponse> DaliModelInstance::CreateResponses(
...
   auto inputs_info = GenerateInputs(requests);
+  tr_gi.stop();
+
+  TimeRange tr_run("[DALI BE] Run processing", 0x488FB1);
```
JanuszL, line 121: You can use predefined constants from TimeRange, not magic values.

src/dali_executor/utils/utils.h
szalpal: Done

src/dali_model_instance.cc
szalpal: Done

src/dali_backend.cc
```diff
@@ -224,6 +224,7 @@ TRITONSERVER_Error* TRITONBACKEND_ModelInstanceFinalize(TRITONBACKEND_ModelInsta
 TRITONSERVER_Error* TRITONBACKEND_ModelInstanceExecute(TRITONBACKEND_ModelInstance* instance,
                                                        TRITONBACKEND_Request** reqs,
                                                        const uint32_t request_count) {
+  TimeRange tr("[DALI BE] ModelInstanceExecute");
```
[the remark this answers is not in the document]
szalpal: I believe it's unnecessary at this stage - they are triggered only once and do not impact the processing too much. At least in DALI case. It might be a good idea to profile these functions as well, however the backend engine in Triton would be a way better place to do that - this would enable the profiling in all other backends as well.

src/dali_model_instance.cc
szalpal: I believe it's not necessary to add one more nested marker here. Almost every line in this function is already covered by nvtx (apart from POD initialization, which should be problematic). Unless I don't see something important?

JanuszL: From my point of view it is easier to read one longer timeline and then check the details, but this is just my opinion. I'm not pushing.

src/dali_backend.cc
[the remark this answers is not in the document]
JanuszL: I agree, that they should be available inside TRITON itself. But we should add and keep them until TRITON adds support for NVTX markers.
While they are supposed to be executed only once I still think it is important to see if they are and how DALI behaves when TIROTN internal changes (maybe not everything looks like we think it does).

src/dali_model_instance.cc
szalpal: In the description I've just pasted the screenshot of how it roughly looks right now. Could you take a look? If there is something I should modify/upgrade, please let me know.

src/dali_backend.cc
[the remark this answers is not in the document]
szalpal: Fair enough, I added those markers

src/dali_model_instance.cc
JanuszL: Looks very nice as it is in the screenshot.

JanuszL (approved):

stiepan (approved):

PR #125, opened 2022-04-07
Autoserialize

szalpal:
This PR adds the `autoserialization` feature to the DALI Backend. 

Experience shows, that it is cumbersome for the user to manually serialize the DALI Pipeline when putting together a model repository. `Autoserialization` allows the user to rely on internal DALI Backend mechanism for serialization instead.

The `autoserialization` feature is documented in the `README.md` file, please refer there for any details.

This PR is related to https://github.com/NVIDIA/DALI/pull/3795


src/dali_model.h
```diff
@@ -86,23 +114,103 @@ class DaliModel : public ::triton::backend::BackendModel {
...
+    do {
+      // Try to load model from the default location
+      LOG_MESSAGE(
+          TRITONSERVER_LOG_VERBOSE,
+          (make_string("Attempting to load DALI pipeline from file: ", default_model_filename)
+               .c_str()));
+
+      // First try to load the serialized model from the default location
+      if (TryLoadModel<FileModelProvider>(model_filename))
+        break;
```
klecki, line 150: This do-while-break is basically `goto`. :upside_down_face: 

And you can basically do it here as the typical `if () return;` and it wouldn't look that weird.

src/model_provider/model_provider.h
```diff
@@ -79,6 +80,71 @@ class FileModelProvider : public ModelProvider {
...
+
+}  // namespace detail
+
+
+class AutoserializeModelProvider : public ModelProvider {
+ public:
+  AutoserializeModelProvider() = default;
+
+  AutoserializeModelProvider(const std::string& module_path, const std::string& target_file) {
+    auto cmd =detail::GenerateAutoserializeCmd(module_path, target_file);
```
klecki, line 133:
```suggestion
    auto cmd = detail::GenerateAutoserializeCmd(module_path, target_file);
```
Nitpick

src/model_provider/model_provider.h
```diff
@@ -79,6 +80,71 @@ class FileModelProvider : public ModelProvider {
...
+  std::stringstream cmd;
+
+  cmd << R"py(python3 -c "
+import importlib, sys
+from nvidia.dali._utils.autoserialize import invoke_autoserialize
+spec = importlib.util.spec_from_file_location('autoserialize_mod', ')py" << module_path << R"py(')
+head_module = importlib.util.module_from_spec(spec)
+sys.modules['autoserialize_mod'] = head_module
+spec.loader.exec_module(head_module)
+invoke_autoserialize(head_module, ')py" << target_file_path << R"py(')
```
klecki, line 97: `target_file_path="some_path')\n do_remote_exploit() \n _ = (' ";`

Invoking strings created by users may be problematic, I don't know if it's viable attack scenario - someone already needs to be able to control the path in triton config, but maybe there will be triton as a service and in this case this is not ideal.


Still, we import user provided python, so :man_shrugging:, already a user code execution.

Maybe adding a validation, that both files exist would help at least a bit?

Also, you mentioned limitations of only importing files ending with `.py`. How about creating a temporary file, copying the contents and importing that?

Also, I feel like this code might be just a Python file that is invoked instead of doing it inline in C++, but it's only a suggestion).

README.md
```diff
@@ -84,6 +83,45 @@ Here's the whole `config.pbtxt` we use for the `ResizePipeline` example:
...
+
+Proper DALI pipeline definition in Python, together with autoserialization, shall meet the
+following conditions:
+1. Only a ``pipeline_def`` can be decorated with ``autoserialize``.
+2. Only one pipeline definition may be decorated with ``autoserialize`` in a given model version.
+
+While loading a model file, DALI Backend follows the precedence:
+1. First, DALI Backend tries to load a serialized model from the user-specified model location (``model.dali`` by default);
+2. If the previous fails, DALI Backend tries to load and autoserialize a Python pipeline
+definition from the user-specified model location. **Important**: Due to Python limitations, name of such file shall end with `.py`, e.g. `model.py`;
```
klecki, line 121: This looks like a limitation that should be lifted by us.

src/dali_model.h
```diff
@@ -32,6 +33,33 @@
...
+namespace detail {
+
+/**
+ * Given a path to a file, assumes that this file is a Python module
+ * and splits the path into two parts:
+ * 1. Path to the file, without the file included in the path
+ * 2. Name of the Python module, inferred from this file. This is a name of the file, without
+ *    an extensions, e.g. `dali.py` -> `dali`; `model.dali` > `model`; etc.
+ */
+inline std::pair<std::string, std::string> SplitPythonModulePath(
```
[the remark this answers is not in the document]
szalpal: Actually, it's a leftover from my previous attempt to call python. I guess I can get rid of it...

src/dali_model.h
szalpal: Well, why not... ;)

README.md
```diff
@@ -84,6 +83,45 @@ Here's the whole `config.pbtxt` we use for the `ResizePipeline` example:
...
+3. If the previous fails, DALI Backend tries to load and autoserialize a Python pipeline
+definition from the ``dali.py`` file in a given model version.
+
 ## Tips & Tricks:
 1. There's a high chance, that you'll want to use the `ops.ExternalSource` operator to feed the encoded
```
banasraf, line 126: Side note: we should edit this readme. There's no other way of using DALI than using external source, so it's not a high chance really. And also `ops` -> `fn`

README.md
```diff
@@ -84,6 +83,45 @@ Here's the whole `config.pbtxt` we use for the `ResizePipeline` example:
...
+following conditions:
+1. Only a ``pipeline_def`` can be decorated with ``autoserialize``.
+2. Only one pipeline definition may be decorated with ``autoserialize`` in a given model version.
+
+While loading a model file, DALI Backend follows the precedence:
+1. First, DALI Backend tries to load a serialized model from the user-specified model location (``model.dali`` by default);
+2. If the previous fails, DALI Backend tries to load and autoserialize a Python pipeline
+definition from the user-specified model location. **Important**: Due to Python limitations, name of such file shall end with `.py`, e.g. `model.py`;
+3. If the previous fails, DALI Backend tries to load and autoserialize a Python pipeline
+definition from the ``dali.py`` file in a given model version.
```
banasraf, line 123: I assume, that if we have the auto-config feature, we'll force users to name the Python files `dali.py` because that will allow us to determine that that's DALI backend and Python backend model.

README.md
klecki: I mean, we have it by default in `model.dali` and according to this we look there for a non-serialized Python code, but it is error to try to read a file that doesn't end with .py. I'm confused.

README.md
banasraf: But I think we could assume that dali.py file means DALI backend

README.md
szalpal: "Default location" means, that the user specified it in the model configuration:
```
default_model_filename: "mymodel.py"
```
It's just that, by default `default_model_filename: "model.dali"`. Basically the takeaway is: If you have a serialized model, call it `model.dali`. If you have a python definition - call it `dali.py`. Probably I should add such sentence there...

szalpal: Added

src/model_provider/model_provider.h
```diff
@@ -79,6 +80,71 @@ class FileModelProvider : public ModelProvider {
...
+  }
+
+  const std::string& GetModel() const override {
+    return fmp_->GetModel();
+  }
+
+  ~AutoserializeModelProvider() override = default;
+
+ private:
+  std::unique_ptr<FileModelProvider> fmp_;
```
banasraf, line 145: What's the reason for using `unique_ptr` instead of just `FileModelProvider`?

src/dali_executor/dali_pipeline.h
```diff
@@ -167,6 +167,14 @@ class DaliPipeline {
     }
   }
 
+  /**
+   * Checks, whether the string contains a formally valid DALI pipeline.
+   * @return True, if `serialized_pipeline `is a formally valid DALI pipeline.
+   */
+  static bool ValidateDaliPipeline(const std::string& serialized_pipeline) {
+    return 0 == daliIsDeserializable(serialized_pipeline.c_str(), serialized_pipeline.length());
```
banasraf, line 175: Nitpick:
`!daliIsDeserializable(serialized_pipeline.c_str(), serialized_pipeline.length());`

src/model_provider/model_provider.h
szalpal: Good catch, this doesn't need to be a pointer.

src/dali_executor/dali_pipeline.h
szalpal: Since `daliIsDeserializable` returns an `int`, I'd rather go with `== 0`

banasraf (dismissed):

src/model_provider/model_provider.h
```diff
@@ -79,6 +80,73 @@ class FileModelProvider : public ModelProvider {
   std::string model_ = {};
 };
 
+
+namespace detail {
+
+inline std::string GenerateAutoserializeCmd(const std::string& module_path,
```
klecki, line 86: I think we should have some validation wrt to the paths here before we use Python.

src/dali_model.h
```diff
@@ -86,23 +87,108 @@ class DaliModel : public ::triton::backend::BackendModel {
...
+   * @param default_model_filename The full path of the model file, specified by the user in the
+   *                               model configuration.
+   * @param fallback_model_filename The full path of the model file, which is a fallback from the
+   *                                default path, in case it is unavailable. The fallback model
+   *                                may only represent unserialized model (which will be a subject
+   *                                for autoserialization).
+   */
+  void LoadModel(std::string default_model_filename, std::string fallback_model_filename) {
+    std::string& model_filename = default_model_filename;
+    std::string target = make_string(".serialized.model.", timestamp(), ".dali");
```
klecki, line 113: Is it fine to do this in the current working directory, or would a temporary file be more suitable?

klecki (commented):
Some nitpicks, and maybe a general remark.

How about just making a distinction, that already serialized models should be in `.dali` files, and the python ones in `.py` files and we do not call it a Python limitation - just the requirement.

If this is a problem for auto-config, maybe they should be called `model_name.dali.py`?

src/model_provider/model_provider.h
szalpal: What kind of path validation do you have in mind?

banasraf:
> How about just making a distinction, that already serialized models should be in `.dali` files, and the python ones in `.py` files and we do not call it a Python limitation - just the requirement.
> 
> If this is a problem for auto-config, maybe they should be called `model_name.dali.py`?

I agree. Naming python files .py makes the most sense and using dali.py should be enough for the auto-conf to recognize the backend.

src/model_provider/model_provider.h
szalpal: We are aware, that this creates a possibility of an exploit, but it is a convenience feature. We advertise this for development. For deployment it is still better to serialize model manually.

README.md
szalpal: Let's assume `dali.py` :)

szalpal:
> Some nitpicks, and maybe a general remark.
> 
> How about just making a distinction, that already serialized models should be in `.dali` files, and the python ones in `.py` files and we do not call it a Python limitation - just the requirement.
> 
> If this is a problem for auto-config, maybe they should be called `model_name.dali.py`?

Isn't this how it works now? 🤔

README.md
```diff
@@ -84,10 +83,53 @@ Here's the whole `config.pbtxt` we use for the `ResizePipeline` example:
...
+        return data
+
+
+Proper DALI pipeline definition in Python, together with autoserialization, shall meet the
+following conditions:
+1. Only a ``pipeline_def`` can be decorated with ``autoserialize``.
+2. Only one pipeline definition may be decorated with ``autoserialize`` in a given model version.
+
+While loading a model file, DALI Backend follows the precedence:
+1. First, DALI Backend tries to load a serialized model from the user-specified model location (``model.dali`` by default);
```
klecki, line 119:
```suggestion
1. First, DALI Backend tries to load a serialized model from the user-specified model location (``model.dali`` if not provided otherwise);
```

README.md
```diff
@@ -84,10 +83,53 @@ Here's the whole `config.pbtxt` we use for the `ResizePipeline` example:
...
+        data = dali.fn.external_source(device="cpu", name="DALI_INPUT_0")
+        return data
+
+
+Proper DALI pipeline definition in Python, together with autoserialization, shall meet the
+following conditions:
+1. Only a ``pipeline_def`` can be decorated with ``autoserialize``.
+2. Only one pipeline definition may be decorated with ``autoserialize`` in a given model version.
+
+While loading a model file, DALI Backend follows the precedence:
```
klecki, line 118: I think we should just stick with python models ending with `.py` and specify this is our requirement and not a Python limitation.

src/dali_model.h
klecki: I would think that we are just producing garbage in the user's directory, but as you wish.

src/model_provider/model_provider.h
klecki: If we can open the provided paths? Maybe it would be clearer to provide the error that way.

szalpal: It's already checked, here: https://github.com/triton-inference-server/dali_backend/blob/be81dbaf8801667d02a43d8d316bd13dc8898cf9/src/model_provider/model_provider.h#L67

README.md
szalpal: Done

README.md
szalpal: Done

klecki (dismissed):

banasraf (approved):

docker/Dockerfile.devel
```diff
@@ -95,13 +95,13 @@ COPY . .
...
 ARG TRITON_BACKEND_API_VERSION="r21.05"
 
-RUN set -ex && mkdir build_in_ci && cd build_in_ci &&                                 \
-    cmake                                                                             \
-      -D CMAKE_INSTALL_PREFIX=/opt/tritonserver                                       \
-      -D CMAKE_BUILD_TYPE=Debug                                                       \
-      -D TRITON_BACKEND_API_VERSION=${TRITON_BACKEND_API_VERSION}                     \
-      -D TRITON_DALI_SKIP_DOWNLOAD=ON                                                 \
-      .. &&                                                                           \
+RUN mkdir build_in_ci && cd build_in_ci && \
```
klecki, line 98: Why did you remove the error checking from here?

klecki (approved):

PR #133, opened 2022-05-31
Install DALI within conda environment

szalpal:
DALI Backend links to DALI using DALI wheel. Currently, the wheel is downloaded in the CMake (via `pip download`) and unzipped into proper location. For the autoserialize feature (#125), simple downloading is no longer sufficient, as we need to run DALI's Python API from the C++*. The solution is to install the wheel in the target docker image. However, if the DALI wheel would be brutally installed by the root in the `tritonserver` image, there might occur a situation, when dependencies between multiple wheels are incompatible with each other.

This PR proposes, how to install the DALI wheel as the conda environment. The build variables remain the same, functionality is maintained. The PR only adds the wheel download+install. Proper activating conda environment (for the purpose of triggering Python script from C++) will be included in next PR.

**Important:**
- [x] Make sure, that the `conda` (most probably in miniconda flavour) is available in the `server` build environment (i.e. a builder docker image). https://github.com/triton-inference-server/server/pull/4476

\* That's because the autoserialize is based on the Python code. And DALI's C/C++ API aren't able to serialize the pipeline.


szalpal:
Note to reviewers: Yes, I'm aware of missing newlines at the end of files. I'll fix it along with applying review remarks.

docker/Dockerfile.devel
```diff
@@ -56,15 +55,30 @@ RUN pip3 install -U pip
...
     ./bootstrap --parallel=$(grep ^processor /proc/cpuinfo | wc -l) -- -DCMAKE_USE_OPENSSL=OFF && \
     make -j"$(grep ^processor /proc/cpuinfo | wc -l)" install && \
     rm -rf /cmake-${CMAKE_BUILD}
 
+# Miniconda
+ENV PATH="/opt/conda/bin:${PATH}"
+RUN MINICONDA_URL="https://repo.anaconda.com/miniconda/Miniconda3-py38_4.12.0-Linux-x86_64.sh"; \
+    SHA256SUM="3190da6626f86eee8abf1b2fd7a5af492994eb2667357ee4243975cdbb175d7a"; \
+    wget "${MINICONDA_URL}" -O miniconda.sh -q && \
+    echo "${SHA256SUM} miniconda.sh" > shasum && \
```
JanuszL, line 72: Do you need this shasum?

docker/Dockerfile.release
```diff
@@ -56,15 +56,30 @@ RUN pip3 install -U pip
...
+    wget "${MINICONDA_URL}" -O miniconda.sh -q && \
+    echo "${SHA256SUM} miniconda.sh" > shasum && \
+    sh miniconda.sh -b -p /opt/conda && \
+    rm miniconda.sh shasum && \
+    ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
+    echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && \
+    echo "conda activate base" >> ~/.bashrc && \
+    find /opt/conda/ -follow -type f -name '*.a' -delete && \
+    find /opt/conda/ -follow -type f -name '*.js.map' -delete && \
+    /opt/conda/bin/conda clean -afy
```
JanuszL, line 81: What is the purpose of this set of commands?

src/dali_executor/CMakeLists.txt
```diff
@@ -30,25 +30,27 @@ include(${tritondalibackend_SOURCE_DIR}/cmake/dali.cmake)
...
+    configure_file(${tritondalibackend_SOURCE_DIR}/cmake/dalienv.yml.in dalienv.yml @ONLY)
+endif()
+
+add_custom_command(
+    OUTPUT dali_env
     COMMAND
-        unzip -q -o ${CMAKE_CURRENT_BINARY_DIR}/nvidia_dali*.whl -d ${CMAKE_CURRENT_BINARY_DIR}/dali
-    COMMAND touch dali_whl  # So that this command won't be re-run every time
-    COMMENT "Acquiring DALI release"
+        conda env create -f dalienv.yml -p ${CMAKE_CURRENT_BINARY_DIR}/dalienv -q
```
JanuszL, line 42: I wonder if we need this additional level of indirection through dalienv.yml, how about:
```suggestion
        conda create -n dalienv python=3.8 -y && conda install -n dalienv -y pip && conda run -n dalienv pip install --extra-index-url ${DALI_EXTRA_INDEX_URL} ${EXTRA_DOWNLOAD_ARGS} ${DALI_DOWNLOAD_PKG_NAME}
```

src/dali_executor/CMakeLists.txt
```diff
@@ -30,25 +30,27 @@ include(${tritondalibackend_SOURCE_DIR}/cmake/dali.cmake)
...
 )
 
-if (${TRITON_DALI_SKIP_DOWNLOAD})
+if(${TRITON_DALI_SKIP_DOWNLOAD})
     get_dali_paths(DALI_INCLUDE_DIR DALI_LIB_DIR DALI_LIBRARIES)
-else ()
-    set(DALI_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/dali/nvidia/dali/include)
-    set(DALI_LIB_DIR ${CMAKE_CURRENT_BINARY_DIR}/dali/nvidia/dali)
+else()
+    set(DALI_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/dalienv/lib/python3.8/site-packages/nvidia/dali/include)
```
JanuszL, line 50: Maybe we should not hardcode it but something like:
```
conda run -n dalienv python -c "import nvidia.dali.sysconfig; print(nvidia.dali.sysconfig.get_include_dir()"
```
and
```
conda run -n dalienv python -c "import nvidia.dali.sysconfig; print(nvidia.dali.sysconfig.get_lib_dir()"
```

docker/Dockerfile.devel
```diff
@@ -56,15 +55,30 @@ RUN pip3 install -U pip
...
+    sh miniconda.sh -b -p /opt/conda && \
+    rm miniconda.sh shasum && \
+    ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
+    echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && \
+    echo "conda activate base" >> ~/.bashrc && \
```
JanuszL, line 77: How it works with the TRION releases process? Does the TRITON image have conda env?

docker/Dockerfile.release
```diff
@@ -56,15 +56,30 @@ RUN pip3 install -U pip
...
     ./bootstrap --parallel=$(grep ^processor /proc/cpuinfo | wc -l) -- -DCMAKE_USE_OPENSSL=OFF && \
     make -j"$(grep ^processor /proc/cpuinfo | wc -l)" install && \
     rm -rf /cmake-${CMAKE_BUILD}
 
+# Miniconda
+ENV PATH="/opt/conda/bin:${PATH}"
+RUN MINICONDA_URL="https://repo.anaconda.com/miniconda/Miniconda3-py38_4.12.0-Linux-x86_64.sh"; \
+    SHA256SUM="3190da6626f86eee8abf1b2fd7a5af492994eb2667357ee4243975cdbb175d7a"; \
+    wget "${MINICONDA_URL}" -O miniconda.sh -q && \
+    echo "${SHA256SUM} miniconda.sh" > shasum && \
```
prak-nv, line 73: Wouldn't you want something like ``sha256sum -c ./shasum && \`` here to really check sha before running it?

docker/Dockerfile.devel
szalpal: FTR: This whole snippet which installs the Miniconda, I took from `continuumio/miniconda3` docker image:
https://github.com/ContinuumIO/docker-images/blob/master/miniconda3/debian/Dockerfile

I removed some of the unnecessary commands.

And answering your question: I don't suppose I need this, but I figured that it might be nice to check the install script checksum. If you feel that it's unnecessary, I'll remove it.

docker/Dockerfile.release
szalpal: Good question. The first two I believe remove some build/install artifacts. The last one, from the conda help:

```
$ conda clean --help

Remove unused packages and caches.

  -a, --all             Remove index cache, lock files, unused cache packages, and tarballs.
  -f, --force-pkgs-dirs
                        Remove *all* writable package caches. This option is not included with the --all flag. WARNING: This will break environments with packages installed using symlinks back to the package cache.
  -y, --yes             Do not ask for confirmation.
```

src/dali_executor/CMakeLists.txt
szalpal: I believe both would work equally well, both have pros and cons. @prak-nv , what your opinion on this?

src/dali_executor/CMakeLists.txt
szalpal: Good point. I'll change this.

docker/Dockerfile.devel
szalpal: I believe, that Triton does not have conda in the build environment. I've put this requirement in the description of the PR. I won't merge this, unless Triton has conda in the build environment.

docker/Dockerfile.release
szalpal: Good point, added.

src/dali_executor/CMakeLists.txt
prak-nv: @szalpal version seems to be more readable/explicit for me.

JanuszL: I'm fine with both.

docker/Dockerfile.devel
JanuszL: Up to you.

prak-nv, line 72: You'll also need ssh256sum here. Sorry, didn't notice this in previous round

src/dali_executor/CMakeLists.txt
JanuszL: So maybe you should use `execute_process` instead of `add_custom_command` ?
Is there any value in making the creation of the conda env a build step instead of part of the configuration process?

szalpal: The value is that it's conducted once per target build. In a hypothetical scenario, that somebody goes:
```
cmake -D TRITON_DALI_SKIP_DOWNLOAD=OFF ..
make
```
And then, during the development touches any of the `CMakeLists.txt` files, consecutive `make` command won't re-download and re-create the conda env. Which is forbidden - user would need to remove entire `build` directory and run `cmake` from scratch.

szalpal: Tried this one. This also has hard time working, because of `configure_file` command above: we need the configured file to create conda environment. OTOH, `configure_file` happens at build stage, with all its repercussions, e.g. if the target is not linked to any executable/library, the file won't be generated. This makes the commands that rely on `dalienv.yml` no-op in the generation step (because this file doesn't exist yet).

Judging all the tweaks here, I believe it is better to hardcode the library paths. It in an easy error to spot - the build will break if there are no libraries at that location, and an easy error to fix if necessary - if the library paths will change, we just need to hardcode different path.

Also, this is not really a technical point, but my gut (and slight experience with CMake) tells me that the build stage is better suitable for workloads such as downloading dependencies. An argument for this might be, that it happens concurrently - when we go `make -j`, one thread handles the download, while everything else that does not depend on the downloading target might be built simultaneously.

Taking into account all this, I believe it would be better to keep creating the conda environment in the build stage, which (unfortunately) stipulates hardcoding the library paths. But, as said earlier, resolving error that might occur here would be rather easy.

JanuszL: Up to you.

> Also, this is not really a technical point, but my gut (and slight experience with CMake) tells me that the build stage is better suitable for workloads such as downloading dependencies. An argument for this might be, that it happens concurrently - when we go make -j, one thread handles the download, while everything else that does not depend on the downloading target might be built simultaneously.

To make sure that it works as intended you need to add a dependency between the DALI backed binaries and `dali_env`. Otherwise, it will attempt to use includes from the env that is just created in parallel.

JanuszL: ..., ok, I see L63. NVM then.

JanuszL (approved):

prak-nv (approved):

PR #134, opened 2022-06-14
Implment auto-config tools

banasraf:
This PR provides tools to validate and auto-fill configuration of DALI backend.

The configuration in Triton is stored as TritonJson::Value. The implemented Validate* functions check the values stored in JSON and compare them with values passed as arguments - those will be later acquired from DALI pipeline. 
The Autofill* functions, additionally to validating, produce new auto-filled TritonJson that will be used as new model configuration.

Changes to CMake were required to use C++17 and also to make Triton libraries available in test binary.

### DIFF
--- a/docker/Dockerfile.devel
+++ b/docker/Dockerfile.devel
@@ -1,6 +1,6 @@
 # The MIT License (MIT)
 #
-# Copyright (c) 2020-2021 NVIDIA CORPORATION
+# Copyright (c) 2020-2022 NVIDIA CORPORATION
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy of
 # this software and associated documentation files (the "Software"), to deal in
@@ -88,5 +88,7 @@ RUN set -ex && mkdir build_in_ci && cd build_in_ci &&                           
       -D TRITON_DALI_SKIP_DOWNLOAD=ON                                                 \
       .. &&                                                                           \
     make -j"$(grep ^processor /proc/cpuinfo | wc -l)" install
+
+ENV LD_LIBRARY_PATH=/opt/tritonserver/lib:/opt/tritonserver/backends/dali/dali:${LD_LIBRARY_PATH}
 
 WORKDIR /opt/tritonserver
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,6 @@
 # The MIT License (MIT)
 #
-# Copyright (c) 2020-2021 NVIDIA CORPORATION & AFFILIATES
+# Copyright (c) 2020-2022 NVIDIA CORPORATION & AFFILIATES
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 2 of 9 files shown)


docker/Dockerfile.devel
```diff
@@ -89,4 +89,6 @@ RUN set -ex && mkdir build_in_ci && cd build_in_ci &&
       .. &&                                                                           \
     make -j"$(grep ^processor /proc/cpuinfo | wc -l)" install
 
+ENV LD_LIBRARY_PATH=/opt/tritonserver/lib:/opt/tritonserver/backends/dali/dali:${LD_LIBRARY_PATH}
```
JanuszL, line 92: Why it is needed?

src/CMakeLists.txt
```diff
@@ -23,26 +23,54 @@ configure_file(libtriton_dali.ldscript libtriton_dali.ldscript COPYONLY)
...
+target_link_libraries(
+        triton-dali-backend-utils PUBLIC
+        triton-core-serverstub
+        triton-backend-utils
+)
+
+set_target_properties(
+        triton-dali-backend-utils
+        PROPERTIES
+            POSITION_INDEPENDENT_CODE ON
```
JanuszL, line 47: Do you need all these options?
Is it based on any example from the TRITON repo or it was selected based on trials and errors?

docker/Dockerfile.devel
banasraf: It's needed to run unittests binary. It dynamically links Triton and DALI libs which are not in LD_LIBRARY_PATH by default

src/CMakeLists.txt
banasraf: For actual backend lib it is based on Triton example. For those helper libraries it was copied and it's actually not needed, so I removed all of the arguments except for C++17.

awolant (approved):
Looks like it works

src/utils/utils.h
```diff
@@ -68,6 +69,23 @@ inline std::string from_string<std::string>(const std::string& str) {
   return str;
 }
 
+template <typename T>
+std::string vec_to_string(const std::vector<T> &vec, const std::string &lbracket = "{",
+                          const std::string &rbracket = "}", const std::string &delim = ", ") {
+  std::stringstream ss;
+  ss << lbracket.c_str();
```
prak-nv, line 76: .c_str() not needed here

src/utils/utils.h
```diff
@@ -68,6 +69,23 @@ inline std::string from_string<std::string>(const std::string& str) {
...
+  auto it = vec.begin();
+  if (vec.size() > 0) {
+    ss << *it;
+    for (++it; it != vec.end(); ++it) {
+      ss << delim;
+      ss << *it;
+    }
+  }
+  ss << rbracket.c_str();
```
prak-nv, line 85: also

banasraf: done

prak-nv (approved):

PR #145, opened 2022-07-14
Fix outputs auto-config. Add max_batch_size validation

banasraf:
It fixes handling outputs in auto-config implementation. 
It differentiates the behavior regarding inputs and outputs.
inputs:
The inputs in the config file are referenced by they names (they have to match the names given in the pipeline), so they can be freely reordered in the config file.

outputs:
Names of the outputs in the config file don't have to match the output names from the pipeline - they are rather used to override the names of the pipeline outputs. So in their case, order matters and is assumed to match the order of pipeline outputs. 

I extended the docstring of the config methods to contain this information. 

New methods:
`ReadMaxBatchSize` - reads and validates max batch size in the config

`ValidateConfig` - composition of ValidateInputs, ValidateOuptputs and ReadMaxBatchSize

### DIFF
--- a/src/config_tools/config_tools.cc
+++ b/src/config_tools/config_tools.cc
@@ -22,6 +22,8 @@
 
 #include "src/config_tools/config_tools.h"
 
+#include <limits>
+
 namespace triton { namespace backend { namespace dali {
 
 
@@ -216,7 +218,12 @@ void AutofillIOConfig(TritonJson::Value &io_object, const IOConfig &io_config,
   TRITON_CALL(io_object.AssertType(common::TritonJson::ValueType::OBJECT));
   TRITON_CALL(new_io_object.AssertType(common::TritonJson::ValueType::OBJECT));
 
-  new_io_object.AddString("name", io_config.name);
+  std::string name;
+  if (io_object.MemberAsString("name", &name) == TRITONJSON_STATUSSUCCESS) {
+    new_io_object.AddString("name", name);
+  } else {
+    new_io_object.AddString("name", io_config.name);
+  }
 
   std::string new_data_type = AutofillDtypeConfig(io_object, io_config.name, io_config.dtype);
   new_io_object.AddString("data_type", new_data_type);
@@ -232,6 +239,25 @@ void ValidateIOConfig(TritonJson::Value &io_object, const IOConfig &io_config) {
   TRITON_CALL(io_object.AssertType(common::TritonJson::ValueType::OBJECT));
   ValidateDtypeConfig(io_object, io_config.name, io_config.dtype);
   ValidateShapeConfig(io_object, io_config.name, io_config.shape);
+}
+
+
+void ValidateAgainstTooManyInputs(TritonJson::Value &ins, const std::vector<IOConfig> &in_configs) {
+   for (size_t i = 0; i < ins.ArraySize(); ++i) {
+    TritonJson::Value io_object(TritonJson::ValueType::OBJECT);
+    ins.IndexAsObject(i, &io_object);
+    std::string name;
+    if (io_object.MemberAsString("name", &name) != TRITONJSON_STATUSSUCCESS) {
+      throw TritonError::InvalidArg(make_string("Missing name in IO config at position ", i));
+    }
+
+    bool in_present = std::any_of(in_configs.begin(), in_configs.end(),
+                                  [&](const auto &ioc) { return ioc.name == name; });
+    if (!in_present) {
+      throw TritonError::InvalidArg(make_string("Configuration file contains config for ", name,
+                                                " but such input is not present in the pipeline."));
+    }
+  }
 }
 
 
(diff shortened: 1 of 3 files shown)


src/config_tools/config_tools.cc
```diff
@@ -235,57 +242,83 @@ void ValidateIOConfig(TritonJson::Value &io_object, const IOConfig &io_config) {
...
+                                  [&](const auto &ioc) { return ioc.name == name; });
+    if (!in_present) {
+      throw TritonError::InvalidArg(make_string("Configuration file contains config for ", name,
+                                                " but such input is not present in the pipeline."));
+    }
+  }
+}
+
+
+void AutofillInputsConfig(TritonJson::Value &ins, const std::vector<IOConfig> &in_configs,
```
banasraf, line 264: This is copies the implementation from the removed method `AutofillIOsConfig`.

src/config_tools/config_tools.cc
```diff
@@ -235,57 +242,83 @@ void ValidateIOConfig(TritonJson::Value &io_object, const IOConfig &io_config) {
...
-    auto ind = FindObjectByName(ios, io_config.name, &io_object);
-    size_t io_index;
+    ins.IndexAsObject(i, &io_object);
+    std::string name;
+    if (io_object.MemberAsString("name", &name) != TRITONJSON_STATUSSUCCESS) {
+      throw TritonError::InvalidArg(make_string("Missing name in IO config at position ", i));
+    }
+
+    bool in_present = std::any_of(in_configs.begin(), in_configs.end(),
+                                  [&](const auto &ioc) { return ioc.name == name; });
```
szalpal, line 255:
```suggestion
                                  [&name](const auto &ioc) { return ioc.name == name; });
```

src/config_tools/config_tools.cc
```diff
@@ -235,57 +242,83 @@ void ValidateIOConfig(TritonJson::Value &io_object, const IOConfig &io_config) {
...
+    if (!in_present) {
+      throw TritonError::InvalidArg(make_string("Configuration file contains config for ", name,
+                                                " but such input is not present in the pipeline."));
```
szalpal, line 258: I started to wonder, is it necessary to restrict number of inputs this way? I mean, if there are any inputs in the model config, that are not present in the DALI pipeline, we can proceed, at least for now. Eventually, when user feeds the request with the input not present in the DALI pipeline, we can throw error at that stage. This way model config, especially during development, does not have to be modified frequently. What do you think?

src/config_tools/config_tools.h
```diff
@@ -132,25 +134,55 @@ void ValidateIOConfig(TritonJson::Value &io_object, const IOConfig &io_config);
...
                           TritonJson::Value &new_inputs);
 
 
 /**
  * @brief Validates outputs array against provided config values
- * auto outputs auto-configured array to new_ios.
+ * Outputs auto-configured array to new_outputs.
+ *
+ * It's assumed that outputs in the config file have the same order as in the pipeline.
+ * If the config gives an output a name, it overrides the name coming from the pipeline.
```
szalpal, line 153:
```suggestion
 * Outputs auto-configured array to new_outputs.
 *
 * It's assumed that outputs in the config file have the same order as those in the DALI pipeline.
 * If the config sets a name of an output, this name overrides the one specified in the DALI pipeline.
```

src/config_tools/config_tools.h
```diff
@@ -132,25 +134,55 @@ void ValidateIOConfig(TritonJson::Value &io_object, const IOConfig &io_config);
...
  */
 void AutofillOutputsConfig(TritonJson::Value &outputs, const std::vector<IOConfig> &out_configs,
                            TritonJson::Value &new_outputs);
 
 
 /**
- * @brief Validates inputs or outputs array against provided config values.
+ * @brief Validate outputs array against provided config values
+ *
+ * Names of the outputs in the config file may not match names of the outputs in the pipeline.
```
szalpal, line 162:
```suggestion
 * @brief Validate outputs array against provided config values.
 *
 * Names of the outputs in the config file do not need to match names of the outputs in the pipeline.
```
I believe `may not` is more like `they are not allowed`

src/config_tools/config_tools.h
```diff
@@ -38,6 +38,8 @@ struct IOConfig {
   dali_data_type_t dtype;
   std::optional<std::vector<int64_t>> shape;
 
+  IOConfig() = default;
+
```
szalpal, line 42: 👍🏻

awolant (approved):

src/config_tools/config_tools.cc
```diff
@@ -295,4 +328,63 @@ void ValidateIOsConfig(TritonJson::Value &ios, const std::vector<IOConfig> &io_c
...
+    TRITON_CALL(outs.IndexAsObject(i, &out_object));
+    ValidateIOConfig(out_object, out_configs[i]);
+  }
+}
+
+
+int ReadMaxBatchSize(TritonJson::Value &config) {
+  int64_t bs = -1;
+  config.MemberAsInt("max_batch_size", &bs);
+  if (bs > std::numeric_limits<int>::max() || bs < -1) {
```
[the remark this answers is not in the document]
banasraf: All integers in JSON are decoded as int64_t, but our batch size is only an int. So the reading looks as follows: parse as int64 -> check if the number fits in int limits -> cast to int.

src/config_tools/config_tools.cc
banasraf: There is no way for a config with too many inputs to be useful - the request you send is validated against that config, so it has to contain all inputs named in the config file and there's no way to feed those inputs to the pipeline, so we would have to just error out.

src/config_tools/config_tools.cc
```diff
@@ -295,4 +328,63 @@ void ValidateIOsConfig(TritonJson::Value &ios, const std::vector<IOConfig> &io_c
     ValidateIOConfig(io_object, io_config);
   }
 }
+
+
+void ValidateInputs(TritonJson::Value &ins, const std::vector<IOConfig> &in_configs) {
+  TRITON_CALL(ins.AssertType(common::TritonJson::ValueType::ARRAY));
+  ValidateAgainstTooManyInputs(ins, in_configs);
```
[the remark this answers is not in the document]
banasraf: We do. In `ValidateInputs` function there's 
```c++
auto ind = FindObjectByName(ins, in_config.name, &in_object);
if (!ind) {
  throw TritonError::InvalidArg(
    make_string("Missing config for \"", in_config.name, "\" input."));
}
```
It will be triggered if configuration specifies too few inputs.

src/config_tools/config_tools.cc
banasraf: Ok, I see that in fact we could just drop the inputs that or not used by the pipelines. 
Still, I think we should guard against such situation because in normal scenario (not developing or debugging) sending inputs that are not used is wasteful and imposes performance overhead.

szalpal (approved):

szalpal (approved):

PR #161, opened 2022-11-02
Add sequence splitting option

banasraf:

This PR adds support for a new parameter to DALI backend. 
The parameter is "split_outer_dim" and it lets split output of DALI pipeline along the outer axis. 
So if the output is a batch of 2 samples of shapes: 
```
{(2, 300, 300), (3, 300, 300)}
```
will be reshaped to a batch of 5 samples:
```
5x{(300, 300}
```

This can be used to split sequences into sub-sequences (in cooperation with `reshape` operator`) or to split sequences into batch of images.

The usage looks like this in config file:
```pbtxt
parameters [
   {
      key: "split_outer_dim",
      value: {string_value: "OUTPUT1:OUTPUT2"}  # list of outputs to split
   }
]
```

Implementation required to changes:
* in `dali_model_instance.cc` the actual reshaping takes place, for inputs that are specified in the config parameter
* in `dali_model.cc`, in the config validation/auto-fill part, the specified dimensionality is adjusted for outputs that are split.


szalpal:
How about changing the name of the feature? I feel that something around `join` fits more, that `split`. For example:
1. When we have a "reshape" like: `[ (2, 480, 640, 3), (3, 480, 640, 3)]  -->  [ (5, 480, 640, 3) ]`, then what we in fact do is to join the samples along outer-most dimension.
2. Analogously, when we have `[ (5, 480, 640, 3) ]  --> [ (2, 480, 640, 3), (3, 480, 640, 3)]`, we split the batch of samples along outer-most dim.

Looking into `numpy` API, `np.concatenate` looks like something we do here: https://numpy.org/doc/stable/reference/generated/numpy.concatenate.html. There's also `np.flatten`, although it has this specific meaning, that it flattens always to 1-D array.

So how about naming this is the flavour of concatenate? E.g. `cat_outer_dims` or `concat_outer_dims`? I'd vote for the former, but it's just a personal preference, `concat` is also totally fine.

src/dali_executor/utils/utils.h
```diff
@@ -102,6 +102,27 @@ std::vector<TensorListShape<Dims>> split_list_shape(const TensorListShape<Dims>
   return result;
 }
 
+
+inline TensorListShape<> split_outer_dim(const TensorListShape<> &shape) {
+  auto in_ndim = shape.sample_dim();
+  ENFORCE(in_ndim >= 2, "To split outer dimension, batch must be at leat 2-dimensional.");
```
stiepan, line 108: It may not be particularly real life use case, but why not allow scalars after the split?

stiepan (dismissed):

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+                   output_dtype=dali.types.UINT8, output_ndim=5)
+def pipeline():
+  vid = fn.external_source(device='cpu', name='INPUT', ndim=1, dtype=dali.types.UINT8)
+  seq = fn.experimental.decoders.video(vid, device='mixed')
+  seq = fn.resize(seq, resize_x=OUT_WIDTH, resize_y=OUT_HEIGHT)
+
+  seq = fn.pad(seq, axis_names='F', align=FRAMES_PER_SEQUENCE)
+  return fn.reshape(seq, shape=[-1, FRAMES_PER_SEQUENCE, OUT_HEIGHT, OUT_WIDTH, 3], name='OUTPUT')
+```
+You can see that we added two operators at the end of the pipeline - `pad` and `reshape`. Padding samples is needed because we want to split them into sub-sequneces of `FRAMES_PER_SEQUENCE` frames each - that means that we need to ensure that the total number of frames in a sample is divisible by `FRAMES_PER_SEQUENCE`. The result of this operator is adding frames filled with zeroes to the end of the sample to align to that number.
```
szalpal, line 24: There's a lot of `that` word in the second sentence. How about rephrasing?
```suggestion
You can see that we added two operators at the end of the pipeline - `pad` and `reshape`. Since we want to split the sequences into sub-sequneces of `FRAMES_PER_SEQUENCE` frames each, we must ensure that the total number of frames in every sample is divisible by `FRAMES_PER_SEQUENCE`. The Pad operator adds frames filled with zeroes to the end of every sample to align to that number.
```

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+After padding the samples, we use the `reshape` operator to add another dimension to the data, so layout-wise the transformation could be represented as `FHWC` -> `NFHWC`, where `N` dimension represents index of a sub-sequence. An example flow could be described as follows: lets say a sample after resizing has a shape of `(13, 300, 300, 3)` - it's a sequence of 13 frames. Assuming the `FRAMES_PER_SEQUENCE` is 5, the sample is then padded to shape `(15, 300, 300, 3)`. Then, the reshape, without modifying any data, changes it to `(3, 5, 300, 300, 3)` - 3 sub-sequences with 5 frames each.
+
+Reshaping data in a described way deosn't solve our problem completely but we can combine it with an additional model parameter that will split the output tensors and treat each sub-sequence as a separate sample in the batch.
```
szalpal, line 28: > Reshaping data in a described way deosn't solve our problem completely

Could you describe, why such reshaping does not solve the problem? I mean, I know why, but I believe it would be good to put it here.

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The parameter is called `split_along_outer_axis` and can be set in the configuration file as follows:
```
szalpal, line 30:
```suggestion
The parameter is called `split_along_outer_axis` and can be set in the model configuration file as follows:
```

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+  return fn.reshape(seq, shape=[-1, FRAMES_PER_SEQUENCE, OUT_HEIGHT, OUT_WIDTH, 3], name='OUTPUT')
+```
+You can see that we added two operators at the end of the pipeline - `pad` and `reshape`. Padding samples is needed because we want to split them into sub-sequneces of `FRAMES_PER_SEQUENCE` frames each - that means that we need to ensure that the total number of frames in a sample is divisible by `FRAMES_PER_SEQUENCE`. The result of this operator is adding frames filled with zeroes to the end of the sample to align to that number.
+
+After padding the samples, we use the `reshape` operator to add another dimension to the data, so layout-wise the transformation could be represented as `FHWC` -> `NFHWC`, where `N` dimension represents index of a sub-sequence. An example flow could be described as follows: lets say a sample after resizing has a shape of `(13, 300, 300, 3)` - it's a sequence of 13 frames. Assuming the `FRAMES_PER_SEQUENCE` is 5, the sample is then padded to shape `(15, 300, 300, 3)`. Then, the reshape, without modifying any data, changes it to `(3, 5, 300, 300, 3)` - 3 sub-sequences with 5 frames each.
```
szalpal, line 26: How about expressing this transformation using the example shapes in a pipeline. I mean something like:
```python
def pipeline():
  vid = fn.external_source(device='cpu', name='INPUT', ndim=1, dtype=dali.types.UINT8)  # (6, 2859274)
  seq = fn.experimental.decoders.video(vid, device='mixed')  # (6, 300, 300, 3)
  seq = fn.pad(seq, axis_names='F', align=FRAMES_PER_SEQUENCE)  # (10, 300, 300, 3)
  return fn.reshape(seq, shape=[-1, FRAMES_PER_SEQUENCE, OUT_HEIGHT, OUT_WIDTH, 3], name='OUTPUT')  # (2, 5, 300, 300, 3)
```
This might be attached to the example above, or maybe specified in some pseudo code, or a chain of operations? Like `input(6, 2859274) -> decode(6, 300, 300, 3)` etc...

The point here would be to guide the user what are the shape changes across the pipeline.

szalpal: Alternatively, you could specify a concrete "problem" at the top of this document and follow this example throughout. Something for example:
```
Let's say we have a 3 video files, which have different number of frames: 20, 13 and 18. The model we want to use always takes a sequence of 4 frames as an input
```
That's just a suggestion, it may make this description better understandable.

docs/examples/video_decoding/README.md
szalpal, line 28:
```suggestion
Reshaping data in a described way doesn't solve our problem completely, but we can combine it with an additional model parameter that will split the output tensors and treat each sub-sequence as a separate sample in the batch.
```

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Reshaping data in a described way deosn't solve our problem completely but we can combine it with an additional model parameter that will split the output tensors and treat each sub-sequence as a separate sample in the batch.
+
+The parameter is called `split_along_outer_axis` and can be set in the configuration file as follows:
+
+```pbtxt
+parameters [
+  {
+    key: "split_along_outer_axis",
+    value: { string_value: "OUTPUT" }
```
szalpal, line 36:
```suggestion
    value: { string_value: "OUTPUT_1:OUTPUT_2" }
```

szalpal: This brings me a thought, would `{ string_value: "OUTPUT_1:" }` work? What would happen if user provides this syntax?

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+  }
+]
+```
+The value of the parameter (specified as `string_value`) is an output name (or list of names seperated with a colon) that should be transformed. We can see the effect of this parameter on the following example: lets assume the DALI pipeline produced a batch of two samples with the shapes {(1, 5, 300, 300, 3), (2, 5, 300, 300, 3)}, so a sample with 1 sub-sequence and a sample with 2 sub-sequences. The batch produced by our model will then have shape {(5, 300, 300, 3), (5, 300, 300, 3), (5, 300, 300, 3)} - a batch of 3 sequences.
```
szalpal, line 40:
```suggestion
The value of the parameter (specified as `string_value`) is the name of the output (or list of names separated with a colon) that should be transformed. We can see the effect of this parameter on the following example: let's assume the DALI pipeline produced a batch of two samples with the shapes `{(1, 5, 300, 300, 3), (2, 5, 300, 300, 3)}`, so the first sample with 1 sub-sequence and the second sample with 2 sub-sequences. The batch produced by our model will then have shape `{(5, 300, 300, 3), (5, 300, 300, 3), (5, 300, 300, 3)}` - a batch of 3 sequences.
```

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+As you can see, the said parameter helps us also handle video files of different lenghts. Outputting such sequences by the DALI pipeline is not feasible in Triton because batches in Triton need to have uniform shape (every sample in the batch should have the same shape). Using the `reshape` - `"split_along_outer_axis"` combination can transform unevenly shaped sequences into a batch of evenly shaped sub-sequences.
```
szalpal, line 42:
```suggestion
As you can see, the said parameter helps us also handle video files of different lenghts. Outputting such sequences by the DALI pipeline is not feasible in Triton, because batches in Triton need to have a uniform shape (every sample in the batch should have the same shape). Using the `reshape` + `"split_along_outer_axis"` combination can transform unevenly shaped sequences into a batch of evenly shaped sub-sequences.
```

szalpal (commented):
Put some comments to the `REAMDE.md`. Mostly they are suggestions, feel free to disregard those.

One feeling I have is that there is no "formal" definition of `split_along_outer_axis`. There is a really good example which explains the simple usage, but I feel that this might be insufficient. Would you envision adding a "proper" documentation, not just an example of this parameter?

docs/examples/video_decoding/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+@autoserialize
+@dali.pipeline_def(batch_size=256, num_threads=4, device_id=0,
+                   output_dtype=dali.types.UINT8, output_ndim=5)
+def pipeline():
+  vid = fn.external_source(device='cpu', name='INPUT', ndim=1, dtype=dali.types.UINT8)
+  seq = fn.experimental.decoders.video(vid, device='mixed')
+  seq = fn.resize(seq, resize_x=OUT_WIDTH, resize_y=OUT_HEIGHT)
+
+  seq = fn.pad(seq, axis_names='F', align=FRAMES_PER_SEQUENCE)
```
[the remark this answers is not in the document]
banasraf: Resize is kind of necessary, because we need to know frame size to perform the reshape.
It wouldn't be needed if we knew that all of our files have some known resolution

docs/examples/video_decoding/README.md
banasraf: I wanted the config file to match the pipeline, so I included only single output. 

Yup, "OUTPUT_1:" would work

docs/examples/video_decoding/README.md
banasraf: done, I think

docs/examples/video_decoding/README.md
banasraf: I did it as python snippet interleaved with comments describing the changes of sample shape

docs/config.md
```diff
@@ -0,0 +1,30 @@
+# DALI model configuration file
+
+## Model parameters
+
+### `split_along_outer_axis`
+This parameters is used to split samples produced by a DALI pipeline removing their outer dimension.
```
szalpal, line 6: Very nice definition :)

src/dali_executor/utils/utils.h
```diff
@@ -102,6 +102,27 @@ std::vector<TensorListShape<Dims>> split_list_shape(const TensorListShape<Dims>
   return result;
 }
 
+
+inline TensorListShape<> split_outer_dim(const TensorListShape<> &shape) {
+  auto in_ndim = shape.sample_dim();
```
szalpal, line 107: How about adding a test for this function?

szalpal (dismissed):

src/dali_executor/utils/utils.h
banasraf: Done

docs/config.md
banasraf: Done

szalpal (approved):

PR #169, opened 2023-01-17
Add ProtoBuf txt parsing to read max_batch_size

banasraf:
When config is read by the tritonserver, there is no way to determine if max_batch_size was set to 0 by the user or was it just not provided. 
To workaround it, I manually load the config text file (if it exists) and parse it (to a limited level) to read the max_batch_size field. 

### DIFF
--- a/src/config_tools/config_tools.cc
+++ b/src/config_tools/config_tools.cc
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2022 NVIDIA CORPORATION & AFFILIATES
+// Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


src/config_tools/config_tools.h
```diff
@@ -178,16 +178,18 @@ void ValidateInputs(TritonJson::Value &ins, const std::vector<IOConfig> &in_conf
...
-int ReadMaxBatchSize(TritonJson::Value &config);
+void ValidateConfig(TritonJson::Value &config, const std::vector<IOConfig> &in_configs,
+                    const std::vector<IOConfig> &out_configs);
 
 
 /**
- * @brief Validate the model max batch size and inputs and outputs configs against provided values.
+ * @brief Read max_batch_size set in a config represented as protocol buffer text format
+ * Returns a max_batch_size value or an empty optional when
+ * the is absent or cannot me parsed as int
```
stiepan, line 190:
```suggestion
 * the value is absent or cannot be parsed as int
```

mzient:
Why not use RapidJSON or some other JSON parser?

src/config_tools/config_tools.cc
```diff
@@ -421,4 +408,121 @@ void ValidateConfig(TritonJson::Value &config, const std::vector<IOConfig> &in_c
...
+  if (text.empty()) return;
+  size_t open_bracket = 0;
+  do {
+    if (text[0] == '\"') {
+      skip_string(text);
+    } else {
+      if (text[0] == bra) ++open_bracket;
+      else if (text[0] == ket) --open_bracket;
+      text.remove_prefix(1);
+    }
```
mzient, line 473: This doesn't handle proper nesting - should be something like:
```suggestion
    if (text[0] == '\"') {
      skip_string(text);
    } else if (text[0] == '[') {
      skip_complex(text, '[', ']');
    } else if (text[0] == '{') {
      skip_complex(text, '{', '}');
    } else {
      text.remove_prefix(1);
    }
```

src/dali_model.h
```diff
@@ -307,6 +313,7 @@ class DaliModel : public ::triton::backend::BackendModel {
   std::vector<IOConfig> pipeline_inputs_{};
   std::vector<IOConfig> pipeline_outputs_{};
   int pipeline_max_batch_size_ = -1;
+  int config_max_batch_size_ = -1;
```
mzient, line 316: Why not `optional<int>` here, too?

mzient (approved):
With a nitpick: use `optional<int>` in the member, too.

banasraf:
> Why not use RapidJSON or some other JSON parser?

This format is not JSON. It is [Protocol Buffer Text Format](https://developers.google.com/protocol-buffers/docs/text-format-spec)

src/config_tools/config_tools.cc
```diff
@@ -421,4 +408,121 @@ void ValidateConfig(TritonJson::Value &config, const std::vector<IOConfig> &in_c
...
+  skip_ignored(text);
+  if (text.empty()) return {};
+  bool negative = false;
+  if (text[0] == '-') {
+    negative = true;
+    text.remove_prefix(1);
+    skip_ignored(text);
+  }
+  size_t end = 0;
+  while (end < text.size() && !(is_whitespace(text[end]) || is_sep(text[end]))) ++end;
```
stiepan, line 489: Why `is_sep`?

stiepan (approved):

PR #174, opened 2023-02-28
Fix input device check

banasraf:


mzient (dismissed):

szalpal:
All tests in pipeline #7454563 passed, but the `finalize` job failed (which reports the result to github). Merging.

PR #180, opened 2023-03-27
DALI version update to 1.24

szalpal:
This is a regular update of DALI version. This time, due to change in DALI's C API, `DaliPipeline` needs to be adjusted to accommodate this change.


src/dali_executor/dali_pipeline.h
```diff
@@ -226,14 +228,14 @@ class DaliPipeline {
   void CreatePipeline() {
     daliCreatePipeline(&handle_, serialized_pipeline_.c_str(), serialized_pipeline_.length(),
                        max_batch_size_, num_threads_, device_id_, 0, 1, 0, 0, 0);
-    assert(handle_.pipe != nullptr && handle_.ws != nullptr);
+    dali_pipeline_created_ = true;
```
mzient, line 231: This isn't necessary.

src/dali_executor/dali_pipeline.h
```diff
@@ -226,14 +228,14 @@ class DaliPipeline {
...
   }
 
 
   void ReleasePipeline() {
-    if (handle_.pipe && handle_.ws) {
-      daliDeletePipeline(&handle_);
-    }
+    if (!dali_pipeline_created_) return;
+    daliDeletePipeline(&handle_);
```
mzient, line 237:
```suggestion
    if (handle_) {
      daliDeletePipeline(&handle_);
      handle_ = nullptr;
   }
```

src/dali_executor/dali_pipeline.h
```diff
@@ -226,14 +228,14 @@ class DaliPipeline {
...
+    dali_pipeline_created_ = false;
```
mzient, line 238:
```suggestion
```suggestion
    if (handle_) {
      daliDeletePipeline(&handle_);
      handle_ = nullptr;
   }
```

src/dali_executor/dali_pipeline.h
```diff
@@ -267,6 +269,7 @@ class DaliPipeline {
   daliPipelineHandle handle_{};
   ::cudaStream_t output_stream_ = nullptr;
   static std::once_flag dali_initialized_;
+  bool dali_pipeline_created_ = false;
```
mzient, line 272: Not necessary.

src/dali_executor/dali_pipeline.h
szalpal: Unfortunately, this is necessary. Otherwise this code (which actually happens during initialization) won't work:
```c++
DaliPipeline p1;
DaliPipeline p2;
p2.CreatePipeline();
p2 = std::move(p1);
```

mzient: But why wouldn't it work? Moving p2 to p1 would just swap the handles and now the `p2 ` looks as if it wasn't ever initialized (it will have a null handle).

JanuszL (approved):

klecki (approved):

mzient (approved):

PR #184, opened 2023-04-13
Add logs to video_input_decoupled test

szalpal:


qa/L0_video_input_decoupled/client.py
```diff
@@ -129,6 +130,10 @@ def parse_args():
                 expected_batch = expected_result[i * BATCH_SIZE : min((i+1) * BATCH_SIZE, len(expected_result))]
                 expected_batch = np.asarray(expected_batch)
                 result_data = result.as_numpy('OUTPUT')
+                if not np.allclose(expected_batch, result_data):
+                    print(f"Subiteration {i=} failed. {result_data[0][0][:100]=} vs {expected_batch[0][0][:100]=}")
```
mzient, line 134: Why `[:100]`? What if the difference is in some other part?

PR #187, opened 2023-04-18
Add `"depleted"` operator trace handling

szalpal:
Should the `VideoInput` operator be properly working with `ExternalSource` in the same pipeline, DALI Backend needs to properly handle the `"depleted"` operator trace. This trace, defined in DALI, specifies when a given operator needs more data at the input. Refers to https://github.com/NVIDIA/DALI/pull/4794

### DIFF
--- a/src/dali_executor/dali_executor.cc
+++ b/src/dali_executor/dali_executor.cc
@@ -47,7 +47,7 @@ void DaliExecutor::SetupInputs(const std::vector<IDescr>& inputs) {
   request_id_++;
   for (auto& inp : c_inputs) {
     input_names_.push_back(inp.meta.name);
-    pipeline_.SetInput(inp, {request_id_str});
+    pipeline_.SetInput(inp, {request_id_.str()});
   }
 }
 
--- a/src/dali_executor/dali_executor.cc
+++ b/src/dali_executor/dali_executor.cc
@@ -144,8 +144,8 @@ bool DaliExecutor::IsInputConsumed() {
   for (auto &name : input_names_) {
     auto trace = pipeline_.TryGetOperatorTrace(name, "depleted");
     if (!trace.has_value()) {
-      throw std::runtime_error("DALI Error: 'depleted' trace shall be available for every input operator.");
-      throw std::logic_error(make_string("DALI internal error: \"depleted\" trace not found for input \"" , name ,"\". It must be defined by all input operators."));
+      throw std::logic_error(make_string("DALI internal error: \"depleted\" trace not found for input \"" ,
+                                         name ,"\". It must be defined by all input operators."));
     }
     if (streq(trace->c_str(), "true")) {
       return true;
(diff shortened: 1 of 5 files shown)


jantonguirao (dismissed):

src/dali_executor/dali_executor.cc
```diff
@@ -136,6 +134,32 @@ bool DaliExecutor::IsNoCopy(device_type_t es_device, const IDescr& input) {
           input.buffers[0].device_id == pipeline_.DeviceId());
 }
 
+
+static bool streq(const char * lhs, const char * rhs) {
+  return strcmp(lhs, rhs) == 0;
+}
+
+
+bool DaliExecutor::IsInputConsumed() {
```
mzient, line 143: Why not 
```suggestion
bool DaliExecutor::IsInputDepleted() {
```
?

src/dali_executor/dali_executor.cc
```diff
@@ -136,6 +134,32 @@ bool DaliExecutor::IsNoCopy(device_type_t es_device, const IDescr& input) {
...
+  for (auto &name : input_names_) {
+    auto trace = pipeline_.TryGetOperatorTrace(name, "depleted");
+    if (!trace.has_value()) {
+      throw std::runtime_error("DALI Error: 'depleted' trace shall be available for every input operator.");
```
mzient, line 147: You're not stating requirements, but rather producing diagnostics. Mentioning the specific name of the operator would be nice, too.
```suggestion
      throw std::logic_error("DALI Error: \"depleted\" trace not found for input \"" + name +"\". "
                             "It must be available for every input operator.");
```

src/dali_executor/dali_executor.cc
```diff
@@ -136,6 +134,32 @@ bool DaliExecutor::IsNoCopy(device_type_t es_device, const IDescr& input) {
...
+    }
+    if (streq(trace->c_str(), "true")) {
+      return true;
+    }
+  }
+  for (auto &name : input_names_) {
+    auto trace = pipeline_.TryGetOperatorTrace(name, "next_output_data_id");
+    if (trace.has_value() && std::stoull(*trace) != request_id_) {
```
mzient, line 155: It'd be better to keep the current ID as a string and compare strings.
The code, as is, would say that any non-number matches `request_id_ == 0`.

src/dali_executor/dali_executor.cc
szalpal: Done

src/dali_executor/request_id.h
```diff
@@ -0,0 +1,66 @@
...
+    return request_id_;
+  }
+
+
+  /**
+   * Postfix increment.
+   */
+  T operator++(int) {
+    T old = request_id_;
+    operator++();
```
mzient, line 46: Readability nitpick:
```suggestion
    ++*this;
```

mzient (approved):

jantonguirao (approved):

PR #194, opened 2023-05-12
Remove unnecessary installation of DALI

JanuszL:


src/dali_executor/CMakeLists.txt
```diff
@@ -82,11 +82,6 @@ target_link_libraries(dali_executor PUBLIC
         CUDA::cudart
         ${DALI_LIBRARIES})
 
-install(
-    DIRECTORY ${DALI_LIB_DIR}
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/backends/dali
-)
-
 target_compile_definitions(dali_executor PUBLIC)
 if (NOT ${TRITON_DALI_SKIP_DOWNLOAD})
```
mzient, line 86: I think that the `install` clause form above should be moved to `else` here, not removed.

JanuszL: I think it is correct.
In the `if` clause we assume that DALI is already installed (thus skipping the download phase).
@szalpal - was that your intention?

szalpal: Here's my thinking about what's happening:

If the `if` clause below is executed, we assume that DALI has been downloaded and `add_custom_command` above (L42) created the conda environment. Therefore in this `if` below, `install` takes the `dalienv` directory in which the conda environment is created and copies it into `dali/conda/envs`. With the change above (`src/CMakeLists.txt:93`) I believe that this DALI version will be used.

Analysing, what's going to happen if the `TRITON_DALI_SKIP_DOWNLOAD=ON`, conda environment will not be created - this stems from the `add_library` in L66 (`dali_env` target is conditionally added). With the `install` removed, it's highly probable that DALI libs won't be installed and the built image will not work.

Unfortunately, this configuration (with `TRITON_DALI_SKIP_DOWNLOAD=ON`) is not tested in the CI due to its limitations. `server_with_backend_debug` job only builds the `Dockerfile.debug` configuration (and it does not upload the artifact, since it wasn't necessary). However, I believe building `Dockerfile.debug` manually and running any test would be sufficient.

JanuszL: There was an issue. Fixed in src/CMakeLists.txt

mzient (approved):

stiepan (approved):

PR #218, opened 2023-10-17
Remove setting layout for external input

banasraf:


banasraf:
This change would undo the input-layout fix and coverage merged in #158. Closing this draft.

PR #222, opened 2023-11-06
Fix building DALI with `TRITON_DALI_SKIP_DOWNLOAD=ON`

szalpal:
This PR introduces a fix for building DALI Backend directly within Triton container with `-D TRITON_SKIP_DALI_DOWNLOAD=ON` option passed to the cmake.

When the aforementioned option is passed to `cmake`, `get_dali_paths` function is responsible for discovering the path of the DALI installation in the system. This function uses DALI and Python directly. In case of building within Triton container, the system-defined `python` path is not the proper one. The proper `python` path is the one created by the conda environment, which has a different path than the system-defined `python` binary. This PR allows to specify which `python` should be used inside `get_dali_paths` invocation.

From now on, when DALI is built within Triton docker container, `TRITON_DALI_BUILD_IN_TRITON` option shall be turned `ON` (it is not the default option).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ message(STATUS "Build configuration: " ${CMAKE_BUILD_TYPE})
 # projects included in this one by FetchContent.
 #
 option(TRITON_DALI_SKIP_DOWNLOAD "Don't download DALI. Use the version provided in the system" OFF)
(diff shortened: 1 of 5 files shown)


jantonguirao (approved):

banasraf:
The original conda-based Triton build path was superseded by #245, and this branch no longer fits current main. Closing this draft.

Issue #165, opened 2022-12-20
DALI backend not releasing device memory


appearancefnp:
Hello!

I was excited for the 16-bit TIFF decoding, but there is a bug regarding the DALI backend not releasing memory when model is unloaded. Even when you unload the DALI model and load it again, it will consume more memory than previously. Although it converges to a fixed number, it is very high - 7Gb for batch size 3.

![image](https://user-images.githubusercontent.com/46784963/208664395-1ef2288e-f934-450e-93d7-cb0089095738.png)

In the image, you can see, that when a DALI model is loaded, the memory increases, which is fine. The dip is when the model is unloaded. It doesn't release the memory to the initial minimum. When loading the model again, the memory increases again. I've done this model "reloading" multiple times and after some time the memory growth stops.

DALI version: 1.22.0dev, but I think the problem persists with older versions too.

```python
from nvidia.dali import pipeline_def
import nvidia.dali
import nvidia.dali.fn as fn
import nvidia.dali.types as types

def set_normalization_value(pixel_value):
    condition = pixel_value > 255.0
    neg_condition = condition ^ True
    return condition * 65535.0 + neg_condition * 255.0


@pipeline_def(
    batch_size=3,
    num_threads=1,
    device_id=0,
    output_dtype=[types.FLOAT],
    output_ndim=[4],  # Dimensions of image, not including batch dimension
)
def decode_pipeline():
    images = fn.external_source(device="cpu", name="input_0", dtype=types.UINT8, ndim=1)
    images_2 = fn.external_source(
        device="cpu", name="input_1", dtype=types.UINT8, ndim=1
    )
    images_3 = fn.external_source(
        device="cpu", name="input_2", dtype=types.UINT8, ndim=1
    )
    images = fn.experimental.decoders.image(
        images,
        device="mixed",
        dtype=types.UINT16,
    )
    images_2 = fn.experimental.decoders.image(
        images_2, device="mixed", dtype=types.UINT16
    )
    images_3 = fn.experimental.decoders.image(
        images_3, device="mixed", dtype=types.UINT16
    )

    images = fn.transpose(images, perm=[2, 0, 1])
    images_2 = fn.transpose(images_2, perm=[2, 0, 1])
    images_3 = fn.transpose(images_3, perm=[2, 0, 1])
    images = fn.cast([images], dtype=types.FLOAT)
    images_2 = fn.cast([images_2], dtype=types.FLOAT)
[...]

PR #225, opened 2024-01-30
Add `release_after_unload` backend option

szalpal:
Per requests included in #165 , this PR adds a possibility to release memory used by DALI Backend after unloading the DALI model.

To do so, user shall pass proper option to `tritonserver` binary invocation:
```
tritonserver --backend-config=dali,release_after_unload=true
```


szalpal:
Helpful information for the review:
1. Most of the code is for the test.
2. The test uses command-line `nvidia-smi` way to determine used memory, as from our experience it's the most reliable way.
3. The test is "duplicated": `release_after_unload` and `dont_release_after_unload`. The duplication stems from the CI design - it's the easiest this way.

docs/config.md
```diff
@@ -28,3 +28,23 @@ with each sample having a shape of:
...
+the lifetime of the process. In other words, when DALI has allocated the memory, it is not released even
+after the DALI Pipeline object gets deleted. The release of the memory happens at the process quit [^1].
+
+This property entails that when using Triton, the GPU memory allocated by DALI is not freed, even
+after unloading the DALI model. However, should you like to free the GPU memory when DALI model is
+unloaded, please pass `release_after_unload` option to `tritonserver` invocation:
+
+```bash
+tritonserver --model-repository /models --log-verbose 1 --backend-config=dali,release_after_unload=true
+```
```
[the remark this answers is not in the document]
szalpal: I considered it and while this would be intuitive and I believe nice default behaviour, I'm worried about current use cases. There might be a situation where changing the default action will introduce unexpectedly high launching times. What do you think about this?

src/dali_executor/dali_pipeline.h
```diff
@@ -68,14 +69,18 @@ class DaliPipeline {
...
     ReleasePipeline();
     ReleaseStream();
+    if (release_buffers_on_delete_) {
+      ReleaseBuffers();
+    }
   }
 
   DaliPipeline(const std::string& serialized_pipeline, int max_batch_size, int num_threads,
-               int device_id) :
+               int device_id, bool release_buffers_on_delete) :
```
mzient, line 78: In this form, I foresee problems with poor allocation performance and fighting between pipelines. To name a few:
1. What happens if there are multiple pipelines with this flag? I think we should count them.
2. What happens if there are pipelines with and without this flag?

I'd suggest having a counter as a static member or a global variable and increment it when we create a new pipeline and decrement when we destroy it. Once all pipelines are down, we can call ReleaseUnusedBuffers, but not before.
If there's a mix of pipelines with and without the flag, we have no way of separating their memory pools, so we're looking at two imperfect options:
a) defer the memory release until all pipelines are destroyed - possibly keeping large memory utilization during upcoming requests
b) release immediately, which will leave the memory used by the pipelines without the flag (because it's still used), but may degrade their performance in upcoming requests.

szalpal: Thank you @mzient for very good points.

Regarding problem 1, Triton solves the case. It's not possible to unload subset of instances of a model - either you unload all or none. Nevertheless, I'll add the counter, as a sanity check.

Regarding problem 2, this is a backend option - it's set once per Triton process invocation. User sets it as I presented in description above:
```bash
tritonserver --backend-config=dali,release_after_unload=true
```
This way either all model instances have the option, or none of them. Therefore the situation you described (pipelines with and without the flag) is not possible. Since there's no perfect solution for this problem, I'll refrain from introducing a path for this. Is that OK for you?

src/dali_executor/dali_pipeline.h
```diff
@@ -68,18 +69,25 @@ class DaliPipeline {
   ~DaliPipeline() {
     ReleasePipeline();
     ReleaseStream();
+    instance_counter_--;
+    if (release_buffers_on_delete_ && instance_counter_ == 0) {
```
mzient, line 73:
```suggestion
    if (--instance_counter_-- == 0 && release_buffers_on_delete_) {
```

src/dali_executor/dali_pipeline.h
```diff
@@ -262,10 +274,12 @@ class DaliPipeline {
   int max_batch_size_ = 0;
   int num_threads_ = 0;
   int device_id_ = 0;
+  bool release_buffers_on_delete_ = false;
 
   daliPipelineHandle handle_ = nullptr;
   ::cudaStream_t output_stream_ = nullptr;
   static std::once_flag dali_initialized_;
+  static int instance_counter_;
```
mzient, line 282:
```suggestion
  static std::atomic_int instance_counter_;
```

szalpal: Done

src/dali_executor/dali_pipeline.h
szalpal: Done

mzient (approved):

qa/L0_dont_release_after_unload/scripts/compare_memory_usage.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python
```
banasraf, line 1: How about, instead of creating two separate tests for that, compare memory usage in a single case but twice (mem before loading, after loading and after unloading)

banasraf (approved):

PR #227, opened 2024-02-23
Fix model counting

banasraf:
The instance counting was bugged which caused multiple issues in CI.

Instead of just fixing it, I changed it to counting loaded models, instead of created pipelines because it better fits the semantics of release_after_unload option and might save us from unnecessary memory release, when pipelines are temporarily created and destroyed.

### DIFF
--- a/src/dali_executor/dali_pipeline.h
+++ b/src/dali_executor/dali_pipeline.h
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2020-2023 NVIDIA CORPORATION & AFFILIATES
+// Copyright (c) 2020-2024 NVIDIA CORPORATION & AFFILIATES
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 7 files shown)


src/dali_executor/dali_pipeline.h
```diff
@@ -70,25 +68,18 @@ class DaliPipeline {
   ~DaliPipeline() {
     ReleasePipeline();
     ReleaseStream();
-    bool is_last_instance = --instance_counter_ == 0;
-    if (is_last_instance && release_buffers_on_delete_) {
```
banasraf, line 74: The buggy behaviour is here.

We don't check if the pipeline handle is not null, so we decrease the instance count even in desctructors of null pipelines. 
This, together with the fact that the DaliPipeline object was movable resulted in such a scenario:
```c++
// [instance_counter == 0]
DaliPipeline p(..) // constructor, instance_counter++
auto p2 = std::move(p);

// p2 destructor, instance_counter--
// p destructor, instance_counter--
// [instance_counter == -1]
```

szalpal (approved):

PR #234, opened 2024-03-27
Dependency update 2024-03

szalpal:
Periodic dependency update, March 2024

This PR adds also a convenient script to update the DLFW & Triton containers versions


tools/update_container_version.sh
```diff
@@ -0,0 +1,50 @@
...
+    exit 1
+fi
+
+echo "Updating the version to $1"
+
+# Dockerfile.release
+sed -i "s/ARG TRITON_VERSION=24.02/ARG TRITON_VERSION=$1/g" docker/Dockerfile.release
+
+# Dockerfile.devel
+sed -i "s/ARG TRITON_VERSION=24.02/ARG TRITON_VERSION=$1/g" docker/Dockerfile.devel
```
banasraf, line 46: It only works when the current version is exactly 24.02?

tools/update_container_version.sh
```diff
@@ -0,0 +1,50 @@
...
+is_correct_version_re='^[0-9][0-9]\.[0-9][0-9]$'
+if ! [[ $1 =~ $is_correct_version_re ]]; then
+    echo "Provided version ($1) does not look correct. Please double-check."
+    exit 1
+fi
+
+echo "Updating the version to $1"
+
+# Dockerfile.release
+sed -i "s/ARG TRITON_VERSION=24.02/ARG TRITON_VERSION=$1/g" docker/Dockerfile.release
```
mzient, line 43: Can we not hardcode the current version inside multiple strings? Even if it's hardcoded in the script, it should be a simple variable somewhere at the top.

tools/update_container_version.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Dockerfile.devel
+sed -i "s/ARG TRITON_VERSION=24.02/ARG TRITON_VERSION=$1/g" docker/Dockerfile.devel
+
+# Efficientnet example
+sed -i "s|+ARG FROM_IMAGE_NAME=nvcr.io/nvidia/pytorch:24.02-py3|+ARG FROM_IMAGE_NAME=nvcr.io/nvidia/pytorch:$1-py3|g" docs/examples/efficientnet/0001-Update-requirements-and-add-Dockerfile.bench.patch
```
mzient, line 49: ...and here...

tools/update_container_version.sh
```diff
@@ -0,0 +1,50 @@
...
+sed -i "s|FROM nvcr.io/nvidia/tritonserver:24.02-py3|FROM nvcr.io/nvidia/tritonserver:$1-py3|g" benchmarks/BM_efficientnet/Dockerfile.torch
```
mzient, line 50: ...and here...

tools/update_container_version.sh
mzient: ...and here...

szalpal: Good catch. Fixed

tools/update_container_version.sh
szalpal: Fixed

tools/update_container_version.sh
szalpal: Fixed

tools/update_container_version.sh
szalpal: Fixed

mzient (approved):

banasraf (approved):

PR #247, opened 2024-06-14
Remove MXNet from key visual as it is no longer supported by DALI

JanuszL:


szalpal:
I don't know what's our approach here, but we could consider adding a solid (non-transparent) background to this image, so that it views better in dark-mode:
image

szalpal (dismissed):

JanuszL:
> I don't know what's our approach here, but we could consider adding a solid (non-transparent) background to this image, so that it views better in dark-mode: image

Fixed

szalpal (approved):

jantonguirao (approved):

PR #266, opened 2025-04-17
Add resnet50 ensemble test

banasraf:


qa/L0_resnet50_ensemble/test_client.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def load_image(img_path: str):
+    """
+    Loads image as an encoded array of bytes.
+    This is a typical approach you want to use in DALI backend
+    """
+    with open(img_path, "rb") as f:
+        img = f.read()
+        return np.array(list(img)).astype(np.uint8)
```
JanuszL, line 60:
```suggestion
    return np.fromfile(img_path, dtype=np.uint8)
```

mzient: Absolutely! It's literally orders of magnitude faster.

qa/L0_resnet50_ensemble/test_client.py
```diff
@@ -0,0 +1,180 @@
...
+        yield dataset[data_idx : data_idx + batch_size]
+        data_idx += batch_size
+
+
+def save_byte_image(bytes, size_wh=(299, 299), name_suffix=0):
+    """
+    Utility function, that can be used to save byte array as an image
+    """
+    im = Image.frombytes("RGB", size_wh, bytes, "raw")
+    im.save("result_img_" + str(name_suffix) + ".jpg")
```
mzient, line 118: Unused?

docs/examples/resnet50_trt/model_repository/resnet50_trt/config.pbtxt
```diff
@@ -26,7 +26,7 @@ input [
 {
     name: "input"
     data_type: TYPE_FP32
-    dims: [ 3, -1, -1 ]
+    dims: [ 3, 224, 224 ]
```
szkarpinski, line 29: Is there a reason why it now needs to be specified explicitly? Is it worth adding a comment?

docs/examples/resnet50_trt/setup_resnet50_trt_example.sh
```diff
@@ -25,9 +25,9 @@ mkdir -p model_repository/dali/1
 mkdir -p model_repository/ensemble_dali_resnet50/1
 mkdir -p model_repository/resnet50_trt/1
 
-docker run -it --gpus=all -v $(pwd):/workspace nvcr.io/nvidia/pytorch:21.03-py3 /bin/bash -cx \
+docker run -it --gpus=all -v $(pwd):/workspace nvcr.io/nvidia/pytorch:25.03-py3 /bin/bash -cx \
```
szkarpinski, line 28: Is this version pin needed, or maybe using `latest` would be a better option? Maybe we should document this in the deps update procedure, so that somebody will bump it from time to time?

szkarpinski (approved):
Left two small suggestions, but it's OK as it is

docs/examples/resnet50_trt/setup_resnet50_trt_example.sh
banasraf: This is a script that is not directly run by tests, so it doesn't matter that much. I don't think there's latest tag but I think it's ok if this serves as instruction that is not necessary fully up to date.

PR #299, opened 2026-06-17
Stop using deprecated tritongrpcclient module

banasraf:
We were using a depracted module for GRPC client for Triton. Additionally, semantics of one of the functions changed causing test failure.

### DIFF
--- a/qa/L0_many_versions/many_versions_client.py
+++ b/qa/L0_many_versions/many_versions_client.py
@@ -24,7 +24,7 @@
 import argparse, os, sys
 import numpy as np
 from numpy.random import randint
-import tritongrpcclient
+import tritonclient.grpc as tritongrpcclient
 from PIL import Image
 import math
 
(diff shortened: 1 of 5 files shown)


JanuszL (dismissed):

JanuszL:
@banasraf - `Migrates to tritonclient.grpc; adds exception handling for is_model_ready in the models_not_loaded loop but omits the same guard in the models_loaded loop, leaving that path broken under the new client behaviour.`

mzient (approved):

qa/L0_many_versions/many_versions_client.py
```diff
@@ -71,7 +71,12 @@ def main():
 
     for name, versions in models_not_loaded.items():
         for ver in versions:
-            if triton_client.is_model_ready(name, str(ver)):
+            try:
+                ready = triton_client.is_model_ready(name, str(ver))
+            except Exception:
```
rostan-t: Nitpick: Catching the precise exception type that `is_model_ready` raises would avoid silencing other exceptions that can potentially happen. I don't know if this method can raise for other reasons.

rostan-t (dismissed):

JanuszL (dismissed):

JanuszL (approved):

rostan-t (approved):

PR #304, opened 2026-08-14
Add Python and C++ formatting checks

banasraf:
This change adds a minimal, pinned local formatting and linting workflow for Python and C++ without reformatting the existing tree.

- configure Ruff for Python formatting and high-signal syntax/name checks
- check C++ formatting with the existing repository `.clang-format`
- provide check-only commit hooks and explicit manual fix hooks through pre-commit
- document setup, scope, exclusions, and local commands
- fix the string identity comparison identified in the multi-input QA client

Validation:
- `pre-commit run --from-ref origin/main --to-ref HEAD`
- representative `clang-format` check
- manual Ruff and clang-format fix hooks verified as idempotent


docs/development.md
```diff
@@ -0,0 +1,58 @@
+# Development checks
+
+The repository uses [pre-commit](https://pre-commit.com/) as a versioned runner for Python linting
+and Python/C++ formatting. Install the pinned runner in a virtual environment:
+
+```shell
+python3.10 -m venv .venv
```
mzient: Do we need to be _that_ precise with Python versioning? It may be useful in CI, but a pain for contributors.

banasraf: Good point. pre-commit still requires Python 3.10 or newer, but contributors don’t need the exact python3.10 executable. I changed the example to use python3 and documented the minimum version separately.

mzient (approved):

mdabek-nvidia (approved):

PR #306, opened 2026-08-17
Fix OOB read in config.pbtxt parser for unterminated strings

akhilraj9:
The DALI backend parses config.pbtxt with a small hand-written parser
(config_tools.cc). skip_string() doesn't handle a quoted string that never
closes: it scans for the closing quote, runs to the end of the buffer, then
calls remove_prefix(end + 1) with end + 1 > size(). string_view::remove_prefix
has no bounds check (it does size_ -= n), so the length underflows to ~SIZE_MAX
and the next skip walks off the end of the buffer — ASan reports a
heap-buffer-overflow read. It's reachable from any model's config.pbtxt at load
time (e.g. a value like `name: "unterminated`).

Fix: clamp the amount removed to text.size().

While in the same function I also hit a null deref in ReadMBSFromPBtxt: after
matching the field name it calls skip_ignored and then reads pb_txt[0], but
skip_ignored can empty the view (e.g. `max_batch_size` followed by a comment
with no trailing newline), so pb_txt[0] dereferences a null string_view. Added
an empty-view guard.

Both are covered by new Catch2 cases in config_tools.test.cc. I verified before/
after by compiling the parser standalone under -fsanitize=address,undefined: the
old code reproduces the OOB read and the null deref, the new code is clean and
still parses valid configs correctly.

### DIFF
--- a/src/config_tools/config_tools.cc
+++ b/src/config_tools/config_tools.cc
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES
+// Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -22,6 +22,7 @@
 
 #include "src/config_tools/config_tools.h"
 
+#include <algorithm>
 #include <limits>
 
 namespace triton { namespace backend { namespace dali {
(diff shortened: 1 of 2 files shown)


banasraf (dismissed):
Looks good to me

banasraf:
Changed the comments slightly to fit better into our codebase. Thanks for your contribution.

banasraf (dismissed):

whoisj (dismissed):

akhilraj9:
> Changed the comments slightly to fit better into our codebase. Thanks for your contribution.

Thank you. I see the CI message that build has failed.

JanuszL:
> > Changed the comments slightly to fit better into our codebase. Thanks for your contribution.
> 
> Thank you. I see the CI message that build has failed.

It was a temporal CI issue. Seem to work on CI.

rostan-t (approved):

